PR #55, opened 2026-07-16
Add portable search and platform-aware shell execution

hijera:
## What changed

- add a portable `rg_tool` that validates POSIX extended regular expressions, uses a system `rg` when available, and falls back to a built-in Go file walker and matcher
- keep `grep` as a compatibility alias and give `glob` the same native fallback when ripgrep is unavailable
- add shared host shell detection with `pwsh -> powershell -> cmd` on Windows and `bash -> sh` elsewhere
- bind `run_command` and `api_key_command` to the detected shell instead of always invoking `sh -c`
- append `<os>`, `<arch>`, and `<shell>` in an `<environment_context>` block to every rendered system prompt, including custom prompt templates
- provide shell-specific tool guidance so the model chooses PowerShell/cmd commands on Windows instead of Unix-only utilities
- update tool exposure, schemas, documentation, and cross-platform tests

## Why

The Windows release could start successfully but important agent paths still assumed a Unix userland: `run_command` and credential helpers always spawned `sh`, while file search depended on externally installed `rg`/`grep`. The model also received no explicit OS or shell context, so it could select commands such as `pgrep`, `grep`, or `find` on a clean Windows installation.

This change makes the core inspection and command paths work in a standalone Windows binary and preserves the fast system-ripgrep path on hosts where it is installed. The built-in matcher follows the POSIX ERE operators documented in the [IBM syntax reference](https://www.ibm.com/docs/en/watson-explorer/11.0.1?topic=queries-posix-regular-expression-syntax-examples).

## User and developer impact

- Windows users no longer need Git Bash merely for `run_command` or `api_key_command`.
- Minimal and distroless environments can search files without a sidecar executable.
- Existing `grep` tool calls continue to work.
- Custom prompt templates still receive authoritative runtime platform context.

## Validation

- `go test ./internal/platform ./internal/tools/shell ./internal/config ./internal/agent ./internal/tools/fs ./internal/tools`
- `golangci-lint run ./...`
- `go test -run '^$' -tags='http,scheduler,memory' ./...`

The full `go test ./...` run on Windows is still affected by the pre-existing `TestConcurrentPatchSessionMetaActivitySync` rename race (`Access is denied`); all packages changed by this PR pass.


EvilFreelancer:
Пришлось немного переделать, но теперь норм.

PR #72, opened 2026-07-26
fix(shell): make PowerShell output readable and exit codes accurate on Windows

hijera:
## Summary

`run_command` on Windows handed PowerShell output to the model as raw bytes in the console **OEM code page** (866 on a Russian install). Anything non-ASCII arrived as mojibake, so when a command failed, the error text the model needed in order to fix itself was unreadable and it retried blindly.

Observed in a real session: the agent tried `gh pr create --body "..."`, got back

```
command failed: exit status 1
? ??????? ???????? ????????????: ".
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : TerminatorExpectedAtEndOfString
```

and spent three turns guessing (including a bash heredoc, which is not PowerShell syntax) before working around it by hand.

## What was actually wrong

Investigating the failure separated it into two different things.

**1. Output encoding — a real bug here.** PowerShell and the native tools it spawns write in the console OEM code page; we read those bytes unchanged.

**2. The `ParserError` itself — not a bug here.** The failing command contained a backtick inside a double-quoted PowerShell string, where the backtick *is* the escape character. Typed at an interactive prompt it fails identically. Verified that Go's `exec` to `powershell -Command` round-trip is clean for embedded quotes, multi-line commands, here-strings and trailing backslashes, so there is no double-parsing layer to remove. This half is addressed as guidance, not as an executor change.

**3. Exit-code fidelity — a bug found along the way.** `powershell.exe -Command` collapses every failure to `1`, so `cmd /c exit 3` was reported as `exit status 1`.

## Changes

- **`internal/platform/powershell.go`** (new) — wraps PowerShell commands in a prologue that sets `[Console]::OutputEncoding` and `$OutputEncoding` to UTF-8, and an epilogue that propagates the real exit code (capturing `$?` before the first `if` overwrites it). The assignment is guarded with `try`/`catch` because it throws when no console is attached. **The command text is embedded verbatim** — nothing is escaped, quoted, or stripped, so backticks and quotes keep the meaning PowerShell would give them at a prompt.
- **`internal/platform/output_windows.go`** / **`output_other.go`** (new) — `platform.DecodeOutput` returns valid UTF-8 untouched and otherwise decodes from the console code page via `MultiByteToWideChar`. This fallback is what makes a **parser error** readable: a script that fails to compile never reaches its own prologue, so its message arrives in the legacy code page no matter what the wrapper does. No-op outside Windows.
- **`internal/tools/shell/run.go`**, **`internal/config/providers.go`** — both captured-output paths go through `DecodeOutput`.
- **`internal/tools/shell/run.go`** — the PowerShell tool description now tells the model to single-quote literal text or use a here-string (`@'...'@`), explains that a backtick inside double quotes alters or breaks the value, and notes that bash heredocs are not PowerShell syntax.

Special characters are **passed through, not stripped**. Stripping them would silently corrupt commit messages, PR bodies and regexes, and would not have fixed the unreadable-error problem that caused the retry loop.

### Rejected alternatives

| Approach | Special chars | Errors readable | Exit code |
|---|---|---|---|
| `-Command` (before) | fine | OEM code page | collapsed to 1 |
| `-EncodedCommand` | fine | **stderr becomes CLIXML** `<S S="Error">` | correct |
| temp `.ps1` + `-File` | fine | fine | **lost, always 0** |
| `-Command` + prologue/epilogue | fine | UTF-8 | correct |

`-EncodedCommand` looks like the obvious fix but flips Windows PowerShell 5.1 into minishell serialization, so error records come back as XML — worse than the bug. A temp script with `-File` swallowed exit codes entirely. The chosen approach needs no temp file, BOM handling, `-ExecutionPolicy` flag, or cleanup.

## Linux and macOS

Unchanged by construction: the POSIX shells still receive `-c <command>` byte for byte, and `DecodeOutput` is the identity function. `TestShellCommandPOSIXUnwrapped` asserts this in code rather than leaving it to review.

## Testing

- **New spec** `features/shell_special_characters.feature` with a godog harness in `internal/tools/shell/bdd_special_characters_test.go`, phrased platform-neutrally so it runs on both Windows and Linux. Confirmed red before the fix: the non-ASCII scenario failed with `output "\x90\xe3\xe1???\xa9 ??\xe1\xe2 - em dash ? ??" does not contain "Русский текст — em dash ✅ 中文"`.
- **Unit tests** in `internal/platform` cover verbatim command embedding, the guarded prologue, `$?`-before-branch ordering in the epilogue, code page 866 decoding, and the POSIX no-change guarantee.
- `make test` green on Windows — all 13 tag combinations.
- `go test ./...` on Linux in Docker (`golang:1.25-bookworm`) across the tag matrix. One unrelated failure surfaced under `http,scheduler,memory`: `TestSavePreservesUpdatedAtWhenMessagesAndActivitySeqUnchanged` in `internal/session`, a pre-existing timing flake (a 1.2 s sleep against second-granularity timestamps) that passes 5/5 in isolation and lives in a package this branch does not touch. Not addressed here.
- `golangci-lint run` clean on the changed packages.
- **End-to-end** against the model from the original session (`qwen3.6-35b-a3b`), in an isolated config and workspace: `— done ✅` round-tripped as UTF-8 through the real tool, and when asked for literal backticks the model chose a single-quoted literal on the first attempt, citing the reason the new description teaches. Output came back as `` Expand `agent.qwen.md` to **7 structured blocks** (21 items) `` verbatim, with no retry and no `ParserError`.

### DIFF
--- a/features/shell_special_characters.feature
+++ b/features/shell_special_characters.feature
@@ -1,0 +1,30 @@
+Feature: Shell commands carry special characters verbatim
+  run_command hands a command to the host shell without the shell re-parsing it as a quoted
+  string, so an agent can pass backticks, quotes, asterisks, parentheses and non-ASCII text
+  straight through. On Windows the command reaches PowerShell as a script file rather than a
+  -Command argument, which is what used to fail with a ParserError; on Linux and macOS the
+  POSIX shell is invoked exactly as before. Commands that fail are still reported as failed.
+
+  Scenario: A command printing shell special characters returns them verbatim
+    Given the host shell detected by the agent
+    When I run a command that prints the literal text:
+      """
+      Expand `agent.qwen.md` from 5 points to **7 structured blocks** (21 items)
+      """
+    Then the command succeeds
+    And the output contains that text verbatim
+
+  Scenario: A command printing non-ASCII text returns it intact
+    Given the host shell detected by the agent
+    When I run a command that prints the literal text:
+      """
+      Русский текст — em dash ✅ 中文
+      """
+    Then the command succeeds
+    And the output contains that text verbatim
+
+  Scenario: A command that exits non-zero is reported as failed with its own exit code
+    Given the host shell detected by the agent
+    When I run a command that exits with code 3
+    Then the command is reported as failed
+    And the failure reports exit code 3
--- a/features/shell_special_characters.feature
+++ b/features/shell_special_characters.feature
@@ -1,9 +1,10 @@
 Feature: Shell commands carry special characters verbatim
-  run_command hands a command to the host shell without the shell re-parsing it as a quoted
-  string, so an agent can pass backticks, quotes, asterisks, parentheses and non-ASCII text
-  straight through. On Windows the command reaches PowerShell as a script file rather than a
-  -Command argument, which is what used to fail with a ParserError; on Linux and macOS the
-  POSIX shell is invoked exactly as before. Commands that fail are still reported as failed.
+  run_command embeds the command text untouched - nothing is escaped, quoted, or stripped -
+  so backticks, quotes, asterisks, parentheses and non-ASCII text keep the meaning the host
+  shell would give them at a prompt. On Windows the command is wrapped in a prologue that
+  switches the output encoding to UTF-8, so results no longer come back in the console OEM
+  code page, and an epilogue that reports the command's own exit code instead of collapsing
+  every failure to 1. On Linux and macOS the POSIX shell is invoked exactly as before.
 
   Scenario: A command printing shell special characters returns them verbatim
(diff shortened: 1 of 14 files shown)


EvilFreelancer:
Одно замечание.

В преамбуле shell_special_characters.feature:4 написано, что команда попадает в PowerShell "as a script file rather than a -Command argument", но реализация в shell.go:37 оставила -Command с прологом и эпилогом (вариант с временным .ps1 в PR отвергнут из-за потери exit-кодов).

Короче текст в спеке стоило бы поправить, на работу тестов это не влияет.

PR #78, opened 2026-08-01
fix(background): ask whether the task's process is running, not whether a pid opens

hijera:
## Problem

`platform.ProcessGroupAlive` decides what `background_list` offers as a leftover and what `background_reap` kills with `taskkill /T /F`. The two implementations added in b8482c8 do not mean the same thing:

- **unix** ? `syscall.Kill(-pid, 0)`. Asks whether a **process group led by `pid`** still has members.
- **Windows** ? `os.FindProcess(pid) == nil`, which is `OpenProcess(...)` (`$GOROOT/src/os/exec_windows.go:84`). Asks whether **a process object with this number can be opened**.

That is a different question, in three ways:

1. **A corpse still opens.** A Windows process object outlives its process for as long as anybody holds a handle to it, so a dead task reads as alive.
2. **No group filter.** Opening by number matches *any* live process. Signalling `-pid` only ever matches a group leader, and that requirement is what quietly filters pid reuse out on unix ? Windows had nothing in its place, so a recycled pid gets handed to `taskkill /T /F` as if it were the task.
3. **Access denied ? "dead"** for a live process the run may not open. Plus a leaked handle on every probe, which is itself what keeps a dead pid resolvable (defect 1).

Both of the first two are reproduced on a real Windows machine by the tests in this PR ? before the fix:

```
--- FAIL: TestProcessGroupAliveRejectsAKilledProcessWhoseHandleIsRetained (5.35s)
    ProcessGroupAlive(56364) = true for a killed process whose handle is still open
--- FAIL: TestProcessGroupAliveRejectsARecycledPID (0.21s)
    ProcessGroupAlive(24336) = true for a record started an hour before this process
```

## Fix

`ProcessGroupAlive(pid int, processStartedAt time.Time) bool`.

**Windows** ? when `CommandRunner` launches a child, it reads the process's exact OS creation time with `GetProcessTimes` and persists that private identity in `meta.json` as `process_started_at` (it is deliberately omitted from the public HTTP task row). The survivor probe then opens the recorded pid with `PROCESS_QUERY_LIMITED_INFORMATION|SYNCHRONIZE`, closes every handle, and asks two fail-closed questions:

- `WaitForSingleObject(h, 0)` separates a running process from a retained corpse. It is not `os.Process.Wait`, which would block until a live process exits and hang the probe.
- `GetProcessTimes` must return the exact persisted creation time. There is no tolerance window: a missing identity, an unreadable creation time, or any mismatch returns false, so `background_reap` never authorizes `taskkill /T /F` for an unproven pid.

**Linux/macOS** ? `syscall.Kill(-pid, 0)` remains unchanged. The process identity is deliberately unused there because the probe addresses the isolated process group rather than any process that happens to own the numeric pid.
## Tests

- **`internal/platform`** now covers the cross-platform liveness contract plus the Windows-only retained-corpse, rapid pid-reuse, missing-identity, and unreadable-creation-time cases. The exact genuine creation time is also checked so the fix cannot pass by rejecting every process.
- **`features/background_reap.feature`** is the first executable spec for reaping anything. The "earlier coddy run" is a second *real* pool over the same session bundle that is then simply abandoned ? a real process, a real pid, against the real probe. 3 scenarios, 16 steps.
- **`internal/bgtask`**'s survivor helper went through `exec.Command("sleep", "600")`, which needs a `sleep` binary on `PATH` that a stock Windows box does not have. It now goes through `CommandRunner` (host shell, own process group, like a real task) and returns the start time its callers must record.

The pid-reuse case is deliberately **not** a scenario: it cannot pass on unix, where a live group leader is a live group leader whatever the record claims. It stays a Windows unit test, which is where the repo rules put a case one platform cannot hold.

## Live end-to-end

`examples/httpserver/http_e2e_background_reap.py` (self-contained, registered in `test_httpserver.sh`) boots its own coddy, has the model start a background task, **kills coddy outright** ? no drain, so nothing stops the child ? reboots on the same home, and asks the model to clean up. Liveness is checked with `tasklist`/`kill -0` from outside, not with coddy's own probe, so the harness is not grading itself.

Verified against **`neuraldeep/gpt-oss-120b`**:

```
  started bg_1 as pid 27460
  coddy was killed and the background process outlived it
  model replied: **Reaped 1 orphaned process:** | bg_1 | 27460 | Start-Sleep -Seconds 600 | Killed |
ok http background reap e2e (bg_1, pid 27460, model neuraldeep/gpt-oss-120b)
```

## Gate

- all 12 Go build-tag combinations ? green locally
- UI production build ? green
- `golangci-lint run ./...` ? `0 issues`
?? Generated with [Claude Code](https://claude.com/claude-code)


EvilFreelancer:
Reviewed this, merged `main` in, and pushed three commits on top: [`4494f8e`](https://github.com/coddy-project/coddy-agent/commit/4494f8e) (merge), [`9262ffe`](https://github.com/coddy-project/coddy-agent/commit/9262ffe) (review follow-ups), [`f409c52`](https://github.com/coddy-project/coddy-agent/commit/f409c52) (CI). Reviewed by hand plus two other agents (Codex, Cursor) as independent readers.

## The defect is real

Checked against the Go tree rather than the description. On Windows `os.FindProcess` is `OpenProcess(STANDARD_RIGHTS_READ|PROCESS_QUERY_INFORMATION|SYNCHRONIZE)` ([`os/exec_windows.go:84`](https://go.dev/src/os/exec_windows.go)), and the handle it returns is released through `runtime.AddCleanup`, so not deterministically. Every claim in the PR holds, with one worth stating more precisely: the handle that keeps a corpse resolvable in production is usually **the probe's own**, since a coddy that was killed loses its handles with it. So probing a pid was itself what made a dead pid keep reading as alive for the rest of the run.

The dangerous half is pid reuse. A record read hours after a crash, a Windows pid already handed to something else, `background_list` offering it, `background_reap` running `taskkill /T /F` on a stranger's tree. Worth fixing.

Also verified the identity survives the `meta.json` round trip: FILETIME is 100 ns granularity, RFC3339Nano carries it losslessly, and `Time.Equal` compares the instant rather than the location, so the exact match is sound. `WaitForSingleObject` is the right primitive here — `GetExitCodeProcess` is ambiguous on exit code 259, and `GetProcessTimes`' `lpExitTime` is documented as undefined for a running process.

## Merge

One textual conflict in `docs/background-tasks.md` (the list of Gherkin specs), resolved as the union.

One semantic conflict git did not surface: `main` brought `Pool.Adopt` and `adoptedHandle` (#88), and this PR widens `bgtask.Handle` with `ProcessStartedAt()`. `startForeground` now captures the identity right after `cmd.Start()`, the same way `CommandRunner` does, and the adopted handle returns it. `cmd` holds the process handle at that point, so the pid cannot be recycled underneath the read. Also dropped the `sleepCommand` method wrapper the merge left with no callers.

## Follow-ups pushed

**The window between proving a pid and killing it was still open.** `Survivors` verified `(pid, creation time)`, then `ReapSurvivors` handed the bare number to `taskkill` — the same mistake in a smaller form. `TerminateProcessGroupByPID` now takes the recorded identity: on Windows it re-opens the pid, re-checks the creation time, and **holds that handle open for the whole of `taskkill`**. Windows keeps a pid allocated while any handle to its process object is open, so nothing can inherit the number in between, and `taskkill /T` still resolves the children that name it as their parent. A record that does not describe the process is refused outright. Liveness is deliberately *not* required there: a leader that already exited can still have running children, and those are what reaping is for. Unix takes the parameter and ignores it, as `ProcessGroupAlive` does. Covered by `TestTerminateProcessGroupByPIDRefusesAPidTheRecordDoesNotDescribe`.

**Records from before `process_started_at` existed** carry a pid and no identity, so on Windows they fail closed and quietly stop being reapable. That is the right behaviour — the alternative reintroduces exactly what this PR removes — but it was undocumented. Now stated in `docs/background-tasks.md` under *Upgrading*, and pinned by `TestLoadPersistedLeavesALegacyRecordWithoutAProcessIdentity`.

**Corrected three comments that claimed more than the code delivers.** The retained handle in production is the probe's own, not a leftover of the killed run. `PROCESS_QUERY_LIMITED_INFORMATION` improves cross-integrity access, it does not guarantee it — access is still the target's to refuse. And on unix, reading the group leader's start time would not close the reuse gap, it would open a worse one: the group outlives its leader, and a leader that exited while its children kept running is precisely the survivor this probe exists to find. That is a stronger reason to leave unix alone than the cost argument that was there.

`TestProcessGroupAliveRejectsARecycledPID` renamed to `...RejectsACreationTimeTheRecordDoesNotMatch` — it never recycles a pid, nothing makes Windows do that on demand, so it drives the same decision the other way round. Worth the test naming what it proves.

## CI now compiles and runs the Windows half

Everything behind `//go:build windows` was invisible to CI, which runs on Linux only. That is how the `Handle` break above got through the merge silently, and it would have hidden a defect in this PR's own Windows code just as well. Three gates, cheapest first:

- `make check-windows` — cross-build plus `go vet` over every non-`ui` tag combination, test files included, so a Windows-only test that stops compiling fails like production code;
- `make lint-windows` — `golangci-lint` over the same half;
- a `windows-latest` job running `go test` for `internal/platform`, `internal/bgtask`, `internal/tools/shell` — the packages whose behaviour differs by OS, and the only place the Windows-only tests can run at all. Scope is deliberately narrower than `./...`: a gate that goes red for reasons nobody has looked at is not a gate.

`.claude/rules/testing.md`, `.cursor/rules/testing.mdc` and `AGENTS.md` say when to run them.

**All four jobs are green on the first run** ([run 30900422692](https://github.com/coddy-project/coddy-agent/actions/runs/30900422692)), so `procgroup_windows_test.go` has now actually executed rather than merely compiled.

## Verification

- CI — `Test suite`, `Test suite (Windows)`, `Windows cross-build`, `Lint`, all green;
- `make test` locally — green, 344 packages across all 13 tag combinations;
- `make lint` and `make lint-windows` — 0 issues;
- `go test -race -count=3` over `internal/platform`, `internal/bgtask`, `internal/tools/shell` — no races, no flakes;
- Linux behaviour is unchanged by construction and pinned by test: the probe and the kill run the same syscalls, `meta.json` serialises byte for byte as before when there is no identity to add (`TestASnapshotWithoutAProcessIdentityPersistsAsItAlwaysDid`), and the only new JSON tag on `Snapshot` is `json:"-"`, so the HTTP and ACP task rows did not move.

Not done, worth knowing: `TestProcessGroupAliveIsRepeatable` proves the probe answers consistently, not that it closes its handles — a leaking implementation would still pass 2000 times. Proving that needs `GetProcessHandleCount`, which `x/sys/windows` does not wrap, so it would mean hand-rolling the syscall for a test I cannot run. Left out on purpose.

PR #79, opened 2026-08-01
Fix ACP slash commands and live MCP reload

### LINKED ISSUE #76: Coddy ACP in Zed does not expose slash commands from loaded skills
## Coddy ACP in Zed does not expose slash commands from loaded skills

Environment:

- Coddy: 0.9.48
- Zed: 1.13.1 stable
- OS: Windows 10
- Coddy launched by Zed as custom External Agent:

```json
{
  "agent_servers": {
    "coddy": {
      "type": "custom",
      "command": "C:\\Users\\q\\AppData\\Local\\Programs\\coddy\\coddy.exe",
      "args": [
        "acp",
        "--home",
        "C:\\Users\\q\\.coddy",
        "--config",
        "C:\\Users\\q\\.coddy\\config.yaml"
      ],
      "env": {}
    }
  }
}


Problem:

Zed starts a `coddy` External Agent thread, but slash commands are unavailable.
Typing `/context-router` shows:

```text
/context-router is not a recognized command in coddy.
Available commands for coddy: none
```

Coddy itself does load skills correctly:

```text
coddy skills list
Search roots:
  C:\Users\q\.agents\skills
  C:\Users\q\.coddy\skills
  C:\Users\q\repos\brain-os\projects\zed\.coddy\skills

24 skill(s)
context-router enabled
...
```

So the issue is not skills discovery. It appears to be ACP notification/schema compatibility.

Relevant Zed log:

```text
agent stderr: ... msg="starting ACP server" version=0.9.48
agent stderr: ... msg="session created" id=sess_... cwd=C:\Users\q\repos\brain-os\projects\coddy mode=agent
Received session notification for unknown session: SessionId("sess_...")
Handler errored ... method="session/update" ... Invalid params ...
missing field `currentModeId`
json:
{
  "sessionId": "sess_...",
  "update": {
    "sessionUpdate": "current_mode_update",
    "modeId": "agent"
  }
}
```

Observed in Coddy source:

```go
type ModeUpdate struct {
    SessionUpdate string `json:"sessionUpdate"`
    ModeID        string `json:"modeId"`
}
```

Zed expects `currentModeId`, not `modeId`.

Likely required fixes:

1. Change `current_mode_update` payload from `modeId` to `currentModeId`:

```go
type ModeUpdate struct {
    SessionUpdate string `json:"sessionUpdate"`
    CurrentModeID string `json:"currentModeId"`
}
```

2. Update send sites:

```go
acp.ModeUpdate{
    SessionUpdate: acp.UpdateTypeCurrentModeUpdate,
    CurrentModeID: params.ModeID,
}
```

3. Review `available_commands_update` timing.

Currently `HandleSessionNew` calls `sendAvailableSlashCommands(id, state)` before returning `SessionNewResult`.
In Zed this can arrive before the client has registered the session, producing:

```text
Received session notification for unknown session
```

This may cause Zed to drop the command catalog, leaving `Available commands for coddy: none`.

Architecture recommendation:

- Do not send session notifications from inside `HandleSessionNew` before the `session/new` response is delivered, or include available commands in the session creation result if the target ACP schema supports that.
- Otherwise schedule `available_commands_update` after `session/new` response is written.

Regression tests needed:

- Wire JSON for `current_mode_update` uses `currentModeId`.
- Zed-compatible ACP `session/new` flow exposes `available_commands_update`.
- Loaded skills from `~/.agents/skills` appear in advertised slash commands.

hijera:
## Summary

- send the `session/new` or `session/load` response before `available_commands_update`, allowing ACP clients to register the session and expose skill slash commands
- emit ACP `current_mode_update` notifications with `currentModeId` (matches the ACP schema's `CurrentModeUpdate.current_mode_id`; the prose docs page still shows the obsolete `modeId`)
- reconnect configured MCP servers for active sessions when settings are saved, while preserving ACP-provided per-session MCP connections; the reconnect re-runs the workspace trust gate and never swaps a session mid-turn
- make the per-session prompt turn lock non-blocking on every platform, so a second concurrent turn fails fast with `ErrSessionTurnBusy` instead of parking (prerequisite for the non-blocking reload; see below)
- add Gherkin happy-path coverage plus HTTP and unit regression tests

## Root cause

Slash-command notifications were written while the session creation request was still being handled, before the matching JSON-RPC response reached the client. Some ACP clients discarded that update because the session was not registered yet. Mode updates also used the obsolete `modeId` field. Separately, config replacement updated the in-memory settings but left each active session's configured MCP clients untouched.

## MCP reload: trust and turn safety

The live-reload path went through `TrustGate.Connect` once it was rebased on `main`'s workspace-trust gate (#83): saving settings re-evaluates trust rather than replaying what the session started with, so a project-local `.coddy/mcp.json` the operator never approved stays cold and a withdrawn approval does not come back. A regression scenario in `features/mcp_project_trust.feature` fails if the reload bypasses the gate.

Swapping a session's configured MCP clients **mid-turn** would strand the tool definitions that turn already handed the model, so its next MCP call would resolve to a server that no longer exists. A reload that arrives while a turn holds the exclusive lock is parked on the session and drained the moment the turn releases it; idle sessions still reconnect before the save response completes.

## Turn lock change (prerequisite, not incidental)

`acquireStubTurnLock` now uses `TryLock` and returns `ErrSessionTurnBusy` instead of blocking. This is a prerequisite for the non-blocking reload (which probes each session's lock and must fail fast rather than park a settings save behind a running turn), and it also fixes a latent bug: on Windows the stub path backs **every** session (no flock), so a second concurrent prompt previously blocked silently instead of returning the `409` the HTTP layer already expects everywhere.

## User impact

Skills now appear through slash commands in ACP clients such as Zed, mode changes use the expected wire format, and MCP servers added in Settings become available to the current session (once trusted, and between turns) rather than only after opening a new session.

## Validation

- `go test ./internal/session ./internal/acp ./internal/agent -count=1`
- `go test -tags=http ./external/httpserver -count=1`
- combined-tag runs (`-tags=http,memory,scheduler`) on the affected packages
- `golangci-lint run ./...` (`0 issues`)
- two pre-existing `internal/bgtask` process-group tests fail only on the Windows dev box; they fail identically on a clean `main` and are unrelated to this branch

Fixes #76


EvilFreelancer:
Thanks @hijera - both root causes here are real, and I confirmed each one by reverting that single fix on this branch and watching the matching test go red. The ACP schema does declare `CurrentModeUpdate.current_mode_id` as a required camelCase field, so `modeId` really is what Zed chokes on, and the catalog really did leave before the `session/new` response.

Review turned up three things on top, so I moved the work to #90 rather than pushing to your fork. Your four commits are preserved there as-is for history, with two commits on top:

- the `--session-id` reopen path was still broken. `session/new` delegates to `loadSessionFromDisk`, which wrote the replayed transcript, the plan, and the context usage before returning, so the same defect stayed on the path that restores a conversation. Only the slash command catalog had been moved out;
- a settings save could strip a session of every MCP server. All sessions dial under the single context `ReplaceConfig` builds, so one hanging server consumed the budget and the sessions after it installed an empty dial result over their healthy clients;
- two overlapping saves stranded the newer one, leaving an idle session on the superseded configuration, because the first save released the turn lock raw without draining.

#90 also carries a reopen scenario in `features/acp_session_integration.feature` and unit coverage for the reload edge cases. `make test` is green and `make lint` reports 0 issues there.

Closing this one in favour of #90. Review very welcome over there.

EvilFreelancer:
Superseded by #90.

PR #88, opened 2026-08-03
fix(shell): adopt a foreground command that outlives its timeout

hijera:
## The bug

A foreground `run_command` running `npm run serve` **hangs the turn forever**.

At the deadline `exec.CommandContext` kills only the direct shell. The `node.exe` grandchildren (`npm-cli.js` → `vue-cli-service`) inherited the write end of the OS pipe that `exec` creates because `Stdout` is a `*bytes.Buffer` and not an `*os.File`, so `cmd.Wait`'s `io.Copy` never sees EOF and `cmd.Run()` never returns.

Reproduced live against a Vue CLI project: **68s+ against a 10s deadline**, released only by killing the node processes by hand — and the dev server is left orphaned holding its port. Two follow-on defects in the same run:

- when the timeout *did* surface, `return "", fmt.Errorf(...)` threw away the buffered output, which already contained `App running at: - Local: http://localhost:8081/`. `internal/agent/react.go` shows the model an error's text and discards the result string;
- the model then re-ran the identical command with `timeout_seconds: 60`, starting a second orphan. A loop.

## The change

Killing such a command is the wrong answer anyway: a dev server or watcher is doing exactly what was asked. It is now **handed to the session task pool** instead — the process keeps running in its own group, its output stream is redirected into the task sink with everything captured so far flushed in first, and the tool answers with the task id followed by that output.

```
Command still running after 30s. It was NOT cancelled: it now runs as background task bg_1 (hard timeout 1h).
Follow it with background_output task_id="bg_1", background_wait, or terminate it with background_stop.
Do NOT run this command again with a larger timeout_seconds: it is already running, and a second copy
would fight the first one for the same ports, locks and files.
Output captured so far:

  App running at:
  - Local:   http://localhost:8081/
```

- **`internal/tools/shell/foreground.go`** (new) — `exec.Command` rather than `CommandContext` (the deferred cancel would kill the process we mean to keep), a detached process group, exactly one `cmd.Wait`, and `cmd.WaitDelay` as the backstop for a grandchild holding the pipe.
- **`internal/tools/shell/switchwriter.go`** (new) — `cmd.Stdout` is fixed at `Start`, so the run always writes here and adoption points it at the task sink.
- **`internal/bgtask`** — `Pool.Adopt` shares the single scheduling path with `Start` (one line of `start` differs). Adopted work gets `max_timeout_seconds`, never the foreground limit it just outlived — `resolveTimeoutSeconds` honours an explicit value verbatim, so reusing the expired one would kill the task in milliseconds. New `Spec.StartedAt` keeps the elapsed clock honest.
- **Result is `(string, nil)`, not an error**, and the notice leads because the output ceiling truncates from the end.
- **`platform.SanitizeOutput`** (new) — strips ANSI and collapses `\r` progress redraws **on read**, so a webpack log is legible. `output.log` stays a raw record; sanitising at capture time would corrupt sequences split across writes.
- **`internal/tools/ssh/exec.go`** — the same discarded-output timeout, fixed the same way, with a mutex-guarded sink for the now-concurrent read. No adoption there: SSH offers no process group to own.
- **Prompts** — servers and watchers go straight to `background: true`; the 30s foreground default is stated; never re-run with a larger `timeout_seconds`.

The command **is** terminated, process group and all, in the three cases where nothing can take ownership: the turn was cancelled, no pool is wired, or the pool refused (`tools.background.enabled: false`, session at `max_concurrent`, draining). The answer then names the exact reason.

No new config key — `tools.background.enabled: false` is already the opt-out. No HTTP surface change.

## Tests

`features/foreground_timeout_adoption.feature` — 5 scenarios, harness in `internal/tools/shell/bdd_foreground_timeout_test.go`.

The grandchild scenario is a **real** regression test, not a vacuous one: the pre-fix path (`exec.CommandContext` + `bytes.Buffer`, 3s deadline) never returned on that command and had to be killed at 45s. The POSIX form needs its trailing `; true` — without it the outer shell `exec`s the inner one, no grandchild exists, and the scenario would pass against the broken code.

Unit tests: adopted hard timeout is the pool maximum and not the expired one; the notice survives a 5-line output ceiling; both terminate fallbacks; cancellation is never adopted; `exit status 3` and `(no output)` shapes preserved; the sanitiser; `switchWriter`; `Pool.Adopt` refusing without running the handover callback.

## Verification

`golangci-lint run ./...` — **0 issues**. `go test ./...` and `-tags http` green apart from `TestStopReachesASurvivorByItsRecordedGroup` and `TestReapSurvivorsKillsAndRecordsThem`, which **fail identically on `main`** on this Windows box (POSIX-shaped tests that find `sleep` via Git Bash, while `ProcessGroupAlive` only checks the leader there). Confirmed by stashing the change and re-running. `-race` was unavailable locally (no gcc), and `make test` was not run in full (it builds the UI via npm) — worth a CI pass on both.

End to end against the original failure, with `neuraldeep/gpt-oss-120b` driving a real `coddy http` in a Vue CLI project: the model ran `npm run serve` in the foreground, the tool answered with `bg_1` plus a clean log, and the model replied **"Сервер запустился на порту 8081"** in a single tool call with no retry loop. Task state: `running`, `timeout_seconds: 3600`, `started_at` at the original foreground start. `background_stop` then freed port 8081 with zero `node.exe` left.

## Known limitation

If a command detaches a daemon and the shell exits, `WaitDelay` closes the pipe after 5s and marks the task finished; later daemon output is lost. Strictly better than today's hang-plus-leak, and documented in `docs/background-tasks.md` rather than hidden.


hijera:
Follow-up commit `400fee7`, driven by watching **gpt-oss-120b** work against a real Vue CLI project with the adoption path live.

**The anti-retry wording guarded the wrong thing.** Told only "do not re-run this command with a larger `timeout_seconds`", the model reached for variants of the same work instead:

```
npm install                       -> adopted as bg_3
npm install --engine-strict=false -> timed_out
yarn install                      -> failed
npm install --ignore-engines      -> running
```

Two installs ended up fighting over one `node_modules` and left the tree without its bin shims. The notice now rules out **the work**, not the command string — same command, a variant with different flags, or another tool that does the same job — and says what to do instead. The `agent.md` bullet is widened the same way.

**And the point of keeping the output is stderr.** A dev server reports a busy port or a missing binary there, and that is what the model needs to read. Both streams have always gone into one sink, foreground and pool alike, so `2>&1` is never needed — but the tool description only said "combined stdout and stderr output" in passing. It now says outright not to add `2>&1`, and for PowerShell why: the operator wraps each error line in an `ErrorRecord` and mangles exactly the message that was worth reading. The adoption notice points at `background_output` for the same reason.

New coverage: a scenario for output written to the error stream with no redirection, plus units for the stderr path, the description text, and the three refusals in the notice. `golangci-lint` clean; `internal/tools/...`, `internal/prompts`, `internal/agent` green.

<details>
<summary>Note on the live run that produced this</summary>

The `node_modules` damage above was caused by my own test harness, not by the patch: it hard-stopped a backgrounded `npm install` four seconds after it started, right when npm had cleaned the tree but not yet recreated the shims. Killing that task was also a terminal state, so `notify_on_finish` woke a fresh turn, and the model started improvising installs from there. Worth knowing as an operational hazard of `background_stop` on a package manager — separate from this PR.
</details>

PR #93, opened 2026-08-06
feat(zcode): auto-attach .cursor/rules by glob via SessionStart/PreToolUse hooks

hijera:
## Summary

ZCode, like Codex, loads its `AGENTS.md` chain once per session and has no glob-based rule attachment, so the detailed rules in `.cursor/rules/*.mdc` never reach the model unless it chooses to open them — which is advisory and gets skipped. This adds a workspace-scoped ZCode hook that delivers those rules deterministically, mirroring the existing Codex bridge (`.codex/hooks/attach_rules.py`, `docs/codex-hooks.md`). `.cursor/rules/` stays the single source of truth.

## What fires when

| Event | Matcher | What is injected |
|---|---|---|
| `SessionStart` | `startup\|resume\|clear\|compact` | every rule whose frontmatter has `alwaysApply: true` |
| `PreToolUse` | `^(Edit\|Write\|MultiEdit\|ApplyPatch)$` | rules whose `globs` cover a file in the pending tool call |

For this repo, `architecture` / `code-style` / `testing` / `workflow` arrive at session start; the other six attach on demand. Editing `external/httpserver/server.go` pulls in `api-layer` + `implementation-order`; editing `external/ui/src/` pulls in `ui-spa` + `ui-verification`.

## Changes

- **`.zcode/hooks/attach_rules.py`** — `SessionStart` injects `alwaysApply: true` rules; `PreToolUse` injects rules whose `globs` match the edited files, once per rule per session. **Key difference from the Codex sibling:** edited paths are recovered from the `tool_input` JSON fields (`file_path` / `path` / `notebook_path` / …), not from an embedded `apply_patch` blob, because that's how ZCode hands paths to hooks. Hand-rolled glob→regex (`fnmatch` is unusable — its `*` crosses `/`), per-session dedup under `<tmp>/zcode-coddy-rules/<session_id>.json`, fail-open on any error. `ZCODE_RULES_DIR` / `ZCODE_RULES_STATE_DIR` override for testing.
- **`.zcode/config.json`** — `hooks.enabled: true` (configuration-file hooks are disabled by default); `SessionStart` + `PreToolUse` handlers. Command uses `python` (not `python3` — on Windows `python3` is a no-op Microsoft Store stub) and `${ZCODE_PROJECT_DIR}` for the repo root. `ApplyPatch` is in the matcher because ZCode aliases `Write`/`Edit` ← `ApplyPatch`.
- **`docs/zcode-hooks.md`** — full guide (mirrors `docs/codex-hooks.md`): triggers, how it works, the Windows interpreter note, how to probe by hand, relationship to the Codex hook.
- **`AGENTS.md`** — rename the section to "Codex, Cursor and ZCode rules"; add a ZCode paragraph parallel to the Codex one.
- **`.gitignore`** — ignore local ZCode session state (`.zcode/plans/`, `.zcode/sessions/`) so shared config under `.zcode/` can be versioned without dragging in per-developer transcripts.

## Why a separate script vs. one shared with Codex

`parse_rule`, `glob_to_regex`, per-session state, the `SessionStart`/`PreToolUse` split, and fail-open are identical. The one real divergence is path recovery (JSON fields vs. patch blob), and the two clients' hook config formats differ (Codex `.codex/hooks.json` vs. ZCode `.zcode/config.json` under `hooks.events` with `enabled: true`). Keeping them separate avoids coupling two clients' I/O contracts; the docs call out that shared logic should be kept in sync.

## Validation

No ZCode session needed — the script reads JSON on stdin and writes JSON on stdout. Verified by hand with synthetic payloads:

- `PreToolUse` on `external/httpserver/server.go` → attaches `api-layer` + `implementation-order` ✅
- file outside any glob → empty output ✅
- repeat call, same session → no-op (dedup state) ✅
- `SessionStart` (`startup`) → `architecture` / `code-style` / `testing` / `workflow` ✅
- `SessionStart` (`clear`) → drops dynamic history, reseeds with always-on set ✅
- corrupt stdin, empty stdin, broken `.mdc` → exit 0, no output, never blocks ✅
- `ApplyPatch` alias and nested path field (`tool_input.commands[].file_path`) handled ✅
- `python -m json.tool .zcode/config.json` valid; `py_compile` on the script passes ✅

## Out of scope

- Does not touch `.codex/` (Codex side unchanged).
- Does not touch `.cursor/rules/` (source of truth).
- No Go code/tests — this is outside the Go build (script + JSON + docs), so the pre-commit lint gate is skipped per `scripts/checks.sh`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,8 @@ Thumbs.db
 
 # Codex cross-review local state
 .codex-review/
+
+# ZCode local session state (plans, transcripts). Shared, versioned config
+# (.zcode/config.json, .zcode/hooks/) is added to git explicitly.
+.zcode/plans/
+.zcode/sessions/
--- a/.zcode/config.json
+++ b/.zcode/config.json
@@ -1,0 +1,33 @@
+{
+  "hooks": {
+    "enabled": true,
+    "events": {
+      "SessionStart": [
+        {
+          "matcher": "startup|resume|clear|compact",
+          "hooks": [
+            {
+              "type": "command",
+              "command": "python \"${ZCODE_PROJECT_DIR}/.zcode/hooks/attach_rules.py\"",
+              "timeout": 10,
+              "statusMessage": "Loading always-on project rules"
+            }
+          ]
+        }
+      ],
+      "PreToolUse": [
+        {
+          "matcher": "^(Edit|Write|MultiEdit|ApplyPatch)$",
+          "hooks": [
+            {
+              "type": "command",
+              "command": "python \"${ZCODE_PROJECT_DIR}/.zcode/hooks/attach_rules.py\"",
+              "timeout": 10,
+              "statusMessage": "Attaching rules for edited files"
+            }
+          ]
+        }
+      ]
+    }
+  }
+}
--- a/.zcode/hooks/attach_rules.py
+++ b/.zcode/hooks/attach_rules.py
@@ -1,0 +1,281 @@
+#!/usr/bin/env python3
+"""Attach `.cursor/rules/*.mdc` to ZCode sessions deterministically.
+
+ZCode has no glob-based rule attachment: its instruction file (`AGENTS.md`) is
+resolved once per session from the working directory up to the project root, so
+nested instruction files never load, and there is no equivalent of the `globs`
+field in `.cursor/rules/*.mdc`. Asking the model to "read the relevant rule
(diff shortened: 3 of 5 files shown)


EvilFreelancer:
Тут тесты падают, надо посмотреть в чём дело и пофиксить.

PR #94, opened 2026-08-07
feat(ui): replace native confirm with frosted-glass React ConfirmDialog

hijera:
## Summary

Replaces every `window.confirm` in the SPA with one reusable, polished React confirmation dialog — a frosted-glass popup built on the existing `WorkspaceFolderModal` shell and the `--coddy-glass-panel-*` token family. Also deletes a workaround that existed solely because of the native dialog.

## Why

The native `window.confirm` is unstyled, blocks the JS thread, and (worse) leaves a stray pointer event on the shell backdrop the instant it closes. That last quirk forced a dedicated `sessionDeleteBackdropSuppress` helper + 480 ms deadline + test. A real in-app dialog fixes the UX and makes that workaround dead code.

## Screenshots

Captured from the real `coddy http` build (`TAGS="http ui scheduler"`) driven by headless Chrome over CDP — not mockups. Sources live in `docs/assets/pw-confirm-*.png`.

### Delete chat — the History 🗑 call site

| Before (surface, no dialog) | After — 1280px dark |
|---|---|
|  |  |

| 1280px light | 390px dark |
|---|---|
|  |  |

### Delete scheduler job — the editor sheet call site

Note the interpolated job id wrapping inside the card rather than overflowing it.

| Before (sheet, no dialog) | After — 1280px dark |
|---|---|
|  |  |

1280px light:

### Escape containment — the regression this PR's review pass fixed

`App.tsx` keeps a `window` keydown listener that closes the sessions drawer and the scheduler editor sheet on Escape. Both are the surfaces this dialog opens from, so dismissing the dialog also collapsed them. Same interaction, both states shown **after pressing Escape on the open confirm dialog**:

| Before fix — sheet collapsed too | After fix — only the dialog closed |
|---|---|
|  |  |

No pre-existing surface changed visually: the `styles.css` diff only *adds* `.confirm-dialog*` rules and modifies no existing selector.

## What changed

**New reusable primitive — `external/ui/src/ui/components/`**
- `ConfirmDialog.tsx` — presentational dialog: portal to `document.body`, `role="dialog" aria-modal="true"`, z-index 120 modal layer, backdrop-click + **Escape** to cancel, focus moves to **Cancel** first so an accidental Enter never confirms a delete. `variant: "danger"` (red trash icon) | `"primary"`.
- `useConfirm.tsx` — `ConfirmProvider` + promise-based `useConfirm()` hook, so call sites keep their control flow: `const ok = await confirm({ title, message, confirmLabel, variant })`.
- Mounted once via `<ConfirmProvider>` in `main.tsx`.

**Call-site replacements (3)**
- `App.tsx` `deleteSession` — "Delete draft?" and "Delete chat?" (🗑 in History).
- `SchedulerJobEditorSheet.tsx` `onDelete` — `Delete scheduler job "<id>"?`.

**Removed dead code**
- `sessions/sessionDeleteBackdropSuppress.ts` + its test + all `App.tsx` wiring (import, `useRef`, two `armSessionDeleteBackdropSuppressUntil` calls, the backdrop-close guard). The suppress window only existed to swallow the native-dialog stray click; the React dialog does not produce it.

**Styling — `external/ui/src/styles.css`**
- `.confirm-dialog*` block: `--coddy-glass-panel-*` tokens, `chat-fade-in` entrance, danger/primary/cancel button variants, light-theme overrides, `prefers-reduced-motion` guard.

## Review fixes (commit `df305ba`)

1. **Escape leaked to the handler underneath** — see the before/after pair above. The listener moved to the capture phase with `stopPropagation()` + `stopImmediatePropagation()`, matching what the blocking native `confirm()` did.
2. **A second `confirm()` hung the first caller** — `resolveRef.current` was overwritten, so the superseded promise never settled. It now resolves as cancelled. Dropped the write-only `PendingState.resolve` field.
3. **Focus was not restored** on close (it fell to `<body>`); the dialog now returns focus to the opener, as the native `confirm()` did.
4. **`aria-modal="true"` had no focus trap** — Tab escaped to the page behind the scrim; Tab/Shift+Tab now wrap inside the card.
5. **Long job ids broke the card** — `Delete scheduler job "<id>"?` interpolates one unbroken token into a fixed-width card; title and message got `overflow-wrap: anywhere`.
6. **`variant="primary"` ignored the theme** — colors were hardcoded `rgba(147, 51, 234, …)`, the dark theme's accent, across all 7 themes. Now `color-mix(in srgb, var(--accent) …)`, the same way `.workspace-chip` does it.

## Tests
- `ConfirmDialog.test.tsx` — 14 unit tests (confirm/cancel/Escape/backdrop, variants, `confirming` disabled state, a11y role, plus Escape containment, Tab wrap, and focus restore).
- `useConfirm.test.tsx` — 5 tests (resolves `true`/`false`, Escape, dialog reused across sequential confirms, promise supersession).
- Existing `SchedulerJobEditorSheet.test.tsx` / `.rename.test.tsx` wrapped in `ConfirmProvider` (they render the consumer directly).

## Verification
- `vitest`: **638 passed** (120 files), incl. 19 new tests.
- `tsc --noEmit`: no new errors in touched files (honoured `exactOptionalPropertyTypes`); the pre-existing failures in `Composer.tsx`, `heroTitleWords.ts`, `questionTypes.ts`, and `Markdown.tsx` are unchanged from `main`.
- `make build TAGS="http ui"`: compiles (UI rebuilt; embed targets are gitignored build output).
- Screenshots above, from the running build.

## Also in this branch

`docs(rules)` commit adds a screenshot step to the BDD/TDD flow in `.claude/rules/workflow.md` (mirrored to `.cursor/rules/workflow.mdc`, `.codex/rules.md`, `AGENTS.md`): every UI edit in a PR ships with a screenshot of the surface it changed. `workflow` now also attaches on `external/ui/**` — its `paths:` were Go/docs only, so the rule never loaded on a UI-only change. This PR is the first one held to it.

## Out of scope (intentionally left as-is)
Discard plan, delete MCP server, Codex sign-out, clear finished tasks, and settings-array item removals still fire without confirmation — this PR only replaces the three existing native confirms, per scope.


EvilFreelancer:
Хай, тесты падают, чекни плиз.

hijera:
да, чекаю

hijera:
но это GLM ,надо еще чуть поревьюить))

PR #95, opened 2026-08-08
Add overflow controls for large tool calls

hijera:
## What changed

- add `More…` / `Less` controls to large `write`, `write_file`, and `apply_patch` argument previews
- reuse the shared `PermissionToolPreview` from current `main`
- keep expanded payloads inside a fixed-height, internally scrollable viewport
- restore full saved arguments when session-list previews contain incomplete JSON
- add component and CSS regression coverage and update the UI behavior documentation

## Why

Large write arguments and patches could stretch the entire conversation, forcing users to scroll the page past a single tool call. Reloaded write cards had an additional issue: the 200-character argument preview could not be parsed, so the large-payload controls were unavailable.

## Impact

Large tool calls now behave like the existing capped tool previews: the transcript keeps a stable height, `More…` enables scrolling inside the card, and `Less` restores the clipped state and resets scrolling to the top. Short payloads retain their natural height.

## Visual verification

The baseline uses the same running shared preview with overflow controls disabled, matching the base behavior. All after images come from the rebased commit `f117beb`.

### Wide: before and bounded

| Before: unbounded | After: clipped with `More…` |
| --- | --- |
| Unbounded write and apply_patch previews | Bounded write and apply_patch previews at 1280px |

### Wide: expanded and internally scrolled

Expanded write and apply_patch previews with internal scroll positions

### Narrow: 390px

| Bounded | Expanded and internally scrolled |
| --- | --- |
| Bounded tool previews at 390px | Expanded tool previews at 390px |

## Validation

- UI suite: 642 tests passed
- focused tool-card suite: 44 tests passed
- Vite 8.1.5 production build, `scripts-sync-to-go.mjs`, and `go build -tags "http ui"` passed
- `golangci-lint run ./...`: 0 issues
- browser verification at 1280px and 390px: no horizontal page overflow, fixed preview height, real internal scrolling, and `Less` reset both scroll positions to `0`
- browser console for the verification surface: no warnings or errors
- exercised real large `write` and `apply_patch` calls through NeuralDeep
- GitHub Actions: Lint, Test suite, Test suite (Windows), and Windows cross-build all passed


EvilFreelancer:
Superseded by #107, which carries both commits from this branch unchanged (f117beb, db0c589 - authorship preserved) plus a cross-review hardening commit and a merge of current main.

Live verification found one gap in the flow this PR targets: the permission answer path reloads the transcript three times, each reload overwrote complete argsText with the truncated 200-char argsPreview, and the one-shot recovery fetch cannot fire twice - so after approving a large write, every large preview (and the next permission gate) rendered empty until a full page reload. #107 keeps complete args across reconciles (pickRicherToolArgs), drops the terminal-status gate so mid-tool restores recover too, re-arms the one-shot fetch, and re-measures overflow on the details toggle event. Details, tests, and screenshots are in #107.

Thanks @hijera - the feature itself works well and ships as-is in the superseding PR.

PR #109, opened 2026-08-16
fix(config): keep reasoning_levels auto-detection through a settings save

hijera:
## Summary

A logical model added through **Settings** offered no reasoning selector, even for a model family the backend detects (`valera/qwen3.8-27b`). This PR fixes both halves of that bug, and adds a **Fetch reasoning levels** button so the field can be filled from what the gateway actually detects instead of from memory.

## Root cause

Two independent defects produced the same symptom, and fixing either alone left it in place.

**1. The settings form seeded an empty list.** `defaultForSchema` gave every new logical model `reasoning_levels: []`. The backend correctly reads an explicitly empty list as "hide the reasoning selector", so detection never ran.

**2. The saved YAML re-created that empty list.** `ModelEntry.ReasoningLevels` was a plain `[]string` with a yaml tag carrying no `omitempty`, so `MarshalConfigYAML` wrote `reasoning_levels: []` for a **nil** slice. Every `PUT /coddy/config` rewrites the whole file and reloads it, so the first Settings save turned detection off again — even for a body that omitted the key entirely.

The mirror image of (2) was also broken: `ModelJSON.ReasoningLevels` had `json:"...,omitempty"`, so `GET /coddy/config` **dropped** a hand-written `reasoning_levels: []`. An operator who used the documented opt-out lost it the first time they opened Settings and saved.

### `reasoning_levels` keeps its three states

`ModelEntry.ReasoningLevels` and `ModelJSON.ReasoningLevels` are now `*[]string`. The two structs are converted directly (`ModelJSON(m)`), so they change together.

| Pointer | YAML | Meaning |
|---|---|---|
| `nil` | key omitted | levels auto-detected from the model id |
| `&[]string{}` | `reasoning_levels: []` | reasoning selector hidden for this model |
| `&[]string{…}` | list | exactly these levels are offered |

A plain slice cannot carry that distinction: with `omitempty` the explicit opt-out is erased, without it every auto-detected model is written as `[]` and silently loses detection on the next load. `yaml.v3` decodes `[]` into a non-nil pointer, so both directions now survive a full save → reload cycle.

### `GET /coddy/config/reasoning-levels`

Resolves the levels a logical model id offers with **no** override configured. The id travels as a query parameter because the entry being edited has not been saved yet.

- Detection is model-id based; the provider prefix is looked up in the active config **only** to apply the Codex remap (`minimal` → `none`), so an id whose provider is not saved yet still gets the un-remapped list.
- A saved override is ignored on purpose — the button reports what the *id* offers, not what is currently configured.
- A model with no reasoning family answers `{"ok":true,"levels":[],"detected":false}`. That is the correct answer, not an error.
- A malformed id (the form is mid-edit) answers **200** with `ok:false` and an `error` for inline display. Only a missing `model` parameter is **400**.

### Settings: `ReasoningLevelsField`

Replaces the generic array editor for `models[].reasoning_levels`, which could express only two of the three states — a model added through Settings could never get back to auto-detection.

- A status line names the current state in words.
- **Fetch reasoning levels** fills the list from the endpoint, using the id currently in the form.
- **Use auto-detected** appears whenever the key is present and removes it, so the next save omits `reasoning_levels` again.
- A fetch that detects nothing **leaves the field untouched** rather than writing `[]` — that would hide the selector, the opposite of what the button was asked to do. A failed request reports the error inline and likewise changes nothing.
- Removing the last level by hand is still the way to reach the `[]` opt-out deliberately.

### CLI: `shift+tab` now works for auto-detected models

`cycleReasoning` read the raw `entry.ReasoningLevels` override instead of the resolved list, so cycling did nothing for any model whose levels are detected rather than configured. It now uses `Config.ReasoningLevelsFor`, the same provider-aware list the composer and `GET /v1/models` expose.

## Tests

Twenty tests added.

- **`features/reasoning_levels_fetch.feature`** — happy path, 3 scenarios (Qwen3 tiers, the Codex `none`-in-place-of-`minimal` remap, and a model with no reasoning support), driven by `external/httpserver/bdd_reasoning_levels_test.go`.
- **`internal/config/reasoning_roundtrip_test.go`** — 3 regressions that run through `MarshalConfigYAML` and a reload, which is where the nil/empty distinction was lost. One of them asserts the written file contains no `reasoning_levels` key at all, pinning the root cause directly.
- **`external/httpserver/reasoning_levels_http_test.go`** — 5 edge cases: missing parameter, malformed id, a saved override being ignored, an unsaved provider, and non-GET.
- **`external/ui/src/ui/settings/ReasoningLevelsField.test.tsx`** — 6 cases covering all three states, the empty-detection guard, and the failure path.
- **`SettingsSection.test.tsx`** — the existing add-a-model regression now runs against a harness schema that also carries `stream` (default `true`), so it catches an item factory that strips more than `reasoning_levels`; plus 2 integration cases for the fetch button.

## Validation

Run on `5b62009` (this branch merged with `main`).

| Check | Result |
|---|---|
| `go test` across all 11 tag combinations in `make test` | pass |
| `npm test` (`external/ui`) | 134 files, **742 tests** pass |
| `golangci-lint run ./...` and the `cli` build-tag variant | **0 issues** |
| `go build` for default / `http` / `cli` / `http,ui` / `scheduler,memory` | pass |

`make test` still reports two failures in `internal/tools`: `TestConfigCommitAbortsWhenStagingCannotBeConsumed` and `TestConfigCommitKeepsStagingConsumedWhenRollbackFails`. These are **pre-existing and unrelated** — verified by stashing this branch's changes and re-running them, which reproduces both identically.

### End-to-end check against a running gateway

Built `-tags "http ui"` and drove the real SPA against a live `coddy http`:

- **Fetch** → `low, medium, high` land in the field; request goes to `?model=valera/qwen3.8-27b`.
- `valera/gpt-4o` → "no auto-detected reasoning levels", field untouched.
- **Save all** → the written `config.yaml` contains **zero** occurrences of `reasoning_levels`, and `GET /v1/models` reports `["low","medium","high"]`. This is the original bug, closed.
- Add + remove a level → `reasoning_levels: []` on disk and `GET /v1/models` reports no levels. The opt-out survives a save too.
- `GET /coddy/config/reasoning-levels` live: `qwen3.8-27b` → `low,medium,high`; `codex/gpt-5.5` → `none,low,medium,high`; `gpt-4o` → `[]`; missing `model` → `400`.

## Docs

`external/httpserver/openapi.go` (path + `CoddyReasoningLevelsResponse` schema), `docs/http-api.md`, `docs/ui.md` (functional checklist), `DESIGN.md` (field contract), `docs/config.schema.json` and `docs/config-reference.md`. The last two also gained `gpt-oss*` and `qwen3*`, which were missing from their auto-detection lists.

## Visual check

Screenshots of Settings → Logical models → Reasoning levels, captured from the running `coddy http` build at 1280×800 and 390×844 (2× DPR): the key-absent state, the state after **Fetch**, a model with nothing detected, light theme, the narrow shell, and the Russian locale.


hijera:
## Visual check

Captured from the embedded `coddy http` build at commit `6ee3347`, using a disposable local configuration.

### Russian locale

| Key absent — levels auto-detected | After **Получить уровни ризонинга** |
| --- | --- |
| Logical model with auto-detected reasoning levels | Logical model after fetching low medium high reasoning levels |

**Narrow screen / Russian locale — no levels detected:** the field remains unset and prompts the operator to add an override only if needed.

Mobile logical model with no detected reasoning levels

### English locale

The action reads **Fetch reasoning levels**.

English logical model after fetching reasoning levels

EvilFreelancer:
## Review follow-up: merged with `main`, cross-review findings fixed

Reviewed with a second agent (Codex, `gpt-5.6-sol`) on the branch merged with `main`; the Cursor reviewer could not run (account out of usage on every model, four attempts), so this is Codex plus a manual pass. Codex's verdict on the original diff was **CHANGES_REQUESTED**; the confirmed items are fixed in `7cb91d2`, `bb9027e` and `92f0493` on this branch, each round re-reviewed by Codex until nothing was left. Round 2 found one more case (manual edits during a pending fetch) and round 3 another (a late answer reverting sibling fields); both are fixed and covered by tests. The Codex confirmation of the last commit could not run: the Codex backend has been answering 404 since then (a fresh session fails the same way), so `92f0493` is verified by the regression test, the full Vitest suite and a manual read only. One first-round finding was rejected and later withdrawn by Codex itself: it flagged the embedded `external/ui/app.js` / `styles.css` as stale tracked assets, but those files are gitignored and regenerated by `make build TAGS="http ui"`, so there is nothing to commit there.

### What changed on the branch

- **Merged `origin/main`** (conflict in `SettingsSection.tsx`): `main` now localizes schema-driven settings labels through `schemaFieldLabel` / `schemaFieldDesc`; the model and `reasoning_levels` overrides both go through it, so the Russian locale shows "Уровни рассуждения" from the dictionary instead of the English schema title.
- **Stale fetch answers are dropped** (Codex, medium). `useReasoningLevels` tickets every request; retyping the model id, pressing **Use auto-detected**, or unmounting the field invalidates the ticket, and a late answer resolves to `null` without touching state. Before, a slow answer could land on a different model id, or resurrect a row deleted from the list through the captured array callback in `SettingsArraySection`. Tests: deferred-answer cases for a retyped id, an unmounted field, and a removed row.
- **A manual edit abandons the pending fetch** (Codex round 2, medium). Adding, changing or removing a level while the request was still in flight let the late answer overwrite the operator's newer choice, including a deliberate `[]` opt-out. Manual edits now invalidate the request ticket before writing (`bb9027e`). Tests: Add during a pending fetch keeps `[""]`; removing the last level during a pending fetch keeps `[]`.
- **A late answer writes through the newest `onChange`** (Codex round 3, medium). `SchemaForm` hands the field a callback closed over the model entry of that render; the fetch callback captured the one from the click, so an answer arriving after a sibling field was edited (the **Stream responses** switch, `max_tokens`) rewrote that field back to its old value. The field now keeps the latest callback in a ref (`92f0493`). Test: toggle **Stream responses** while the request is pending, settle, and assert the entry keeps `stream: false` together with the fetched levels.
- **Provider type comes from the form, not from disk** (Codex, medium). The Codex `minimal` -> `none` remap used to be decided by the saved provider, so a provider row changed from `codex` to `openai` but not yet saved fetched `none` and then persisted a wrong explicit override. The field now sends `provider_type` (looked up in the settings document being edited), and `GET /coddy/config/reasoning-levels` honours it over the saved config; without the parameter the old lookup applies. New `config.ReasoningLevelsForProviderType`, `ReasoningLevelsFor` delegates to it. Tests: Go unit table, HTTP cases (unsaved codex provider, saved codex provider switched to openai, blank hint), a new happy-path scenario in `features/reasoning_levels_fetch.feature`, and UI cases asserting the query string.
- **Status line describes the list first** (Codex low, also seen live). After a fetch that found nothing or failed, adding a level by hand kept showing the old message, and retyping the model id kept it too. The list state (override / hidden) now wins, fetch feedback only applies while the key is absent, and a model-id change clears it. Before/after below.
- **OpenAPI error shape** (Codex, low). `400`/`500` on the new route referenced the nested `ErrorEnvelope`, while the handler writes the flat `{"ok":false,"error":...}`; the spec now references `CoddyConfigValidateResponse` for both. The route was also added to the path list `TestOpenAPISpecPathsAndVersion` checks.
- **Nits**: `DetectReasoningLevels` (exported, unused) removed; `ConfigToJSONDTO` / `JSONDTOToConfig` clone `ReasoningLevels` and `Stream` instead of sharing the pointers across the struct conversion, matching the other pointer-typed sections.
- **Repo conventions**: the round-trip tests moved into `internal/config/reasoning_test.go` and the HTTP edge cases into `external/httpserver/reasoning_test.go` (one test file per concern, as the other packages do). The bundled `configure-coddy` skill now documents the three states of `reasoning_levels` and the `delete models[...].reasoning_levels` command; `config.example.yaml` says that omitting the key means auto-detect; `docs/http-api.md`, `docs/ui.md`, `DESIGN.md` describe `provider_type` and the status precedence.

### Validation

| Check | Result |
|---|---|
| `make test` (all tag combinations, incl. `http,ui` after `ui-build`) on `7cb91d2`; `bb9027e` and `92f0493` touch only TS and docs | pass |
| `make lint` (host and `cli` tag), pre-commit on both commits | 0 issues |
| `npm test` in `external/ui` on `92f0493` | 137 files, 835 tests pass |
| `vite build` on `bb9027e` | ok |
| GitHub CI on `7cb91d2` (Lint, Test suite, Test suite Windows, Windows cross-build) | pass |
| GitHub CI on `92f0493` (Lint, Test suite, Test suite Windows, Windows cross-build) | pass |
| `TestConfigCommit*` in `internal/tools`, reported as pre-existing failures in the PR description | pass on current `main` |

### Visual check

Captured from the vite dev build proxied to a running `coddy http` (`-tags http`), dark theme, 1280x800 and 390x844.

**Status line after adding a level by hand following an empty detection, 1280 / English.** Left is the branch as submitted, right is the fix.

| Before | After |
| --- | --- |
| before | after |

**Field states, 1280 / English:** key absent, after **Fetch reasoning levels**, model without a reasoning family.

| Auto-detected | Fetched | Nothing detected |
| --- | --- | --- |
| auto | fetched | none |

**Russian locale, 1280:** the label and description now come from the `settings.schema.models.reasoning_levels.*` dictionary entries (merge with `main`).

| Auto-detected | Manual level after empty detection |
| --- | --- |
| ru auto | ru manual |

**Narrow shell, 390:** English fetched and manual-after-empty, Russian fetched.

| EN fetched | EN manual | RU fetched |
| --- | --- | --- |
| 390 en fetched | 390 en manual | 390 ru fetched |

### Left as is, for the author / maintainer to decide

- Russian copy mixes two terms on one screen: the schema label from `main` says "Уровни рассуждения" while the branch's buttons say "Получить уровни ризонинга" (pinned by a test as requested wording). Worth settling on one.
- `PUT /coddy/config` documents its `400`/`500` with the same nested `ErrorEnvelope` mismatch this branch had; not touched here since it predates the PR.
- `npm test` is not part of `make test` or CI, so the Vitest suites only run locally.

PR #123, opened 2026-08-25
feat(ui): complete Russian localization of conversation surfaces

hijera:
## Summary

Completes the Russian localization of the SPA conversation surfaces. Remaining hardcoded user-facing strings now go through `t()` / `useT()` with English/Russian dictionary parity.

## Follow-up review

- Localized workspace selection, permission/question prompts, stream and offline errors, file-type labels, background tasks, cron descriptions, draft titles, and confirmation-dialog fallback copy.
- Fixed the truncated-output indicator in the background-task panel.
- Fixed stale translated error text after switching the UI language.
- Added regression coverage for the new Russian strings and the truncated-output state.

## Verification

- `cd external/ui && node node_modules/vitest/vitest.mjs run` — 133 files, 739 tests passed.
- UI build and embedded-asset sync passed.
- `go build -tags 'http ui' ./cmd/coddy` passed.
- `make lint` passed (also run by the pre-commit hook).
- Local visual check at 1280x900 and 390x844 with `?lang=ru` passed.

The full Go tag matrix reaches two existing failures in `internal/tools` (`TestConfigCommitAbortsWhenStagingCannotBeConsumed` and `TestConfigCommitKeepsStagingConsumedWhenRollbackFails`). This PR does not change that package.

## Screenshots

Running Vite build, Russian locale:

| Wide — 1280×900 | Narrow — 390×844 |
| --- | --- |
| Russian localized desktop UI | Russian localized mobile UI |

Permission prompts, background-task states, and scheduler/job forms require an active backend session and could not be captured from the standalone local UI; their localized states are covered by unit tests.


hijera:
Pushed two follow-up commits addressing review findings.

## `98e106d` — three localization defects

**Hero title sentinel was a raw NUL byte.** `ChatScreen.tsx` split the translated hero sentence on a marker that was written as a literal `\0` character in the source file. It worked at runtime, but the file became binary to text tooling — `grep` and `git grep` reported `Binary file ... matches` and skipped its contents, so `ChatScreen.tsx` had dropped out of codebase-wide searches. Now written as a `"\u0000"` escape, with a comment explaining why a sentinel is needed instead of splitting on whitespace.

**`save failed ()` in the plan card.** The no-status branch called `translate("prompts.planSaveFailed", { status: "" })` against `"save failed ({status})"`, leaving empty parens. Added `prompts.planSaveFailedNoStatus` for that path.

**Schedule validation lost cronstrue's diagnostics.** `describeCronScheduleOrError` had replaced `e.message` (which names the offending field, e.g. "Expression has only 3 parts") with a generic translated sentence. The original message is shown again; the translated copy is now only the fallback when there is none. Empty input still returns the translated `scheduler.cron.required`.

## `bb02611` — CLDR plural support

Russian needs three forms where English needs two, so `"{count} строк"` rendered as "1 строк" / "2 строк", and `BackgroundTasksChip` worked around the gap with a `count === 1` ternary picking between two keys in the component.

Added `translatePlural` / `tp(key, count)` on top of `Intl.PluralRules`. The dictionary holds one entry per CLDR category under `key.category`; the entry always receives `{count}`; lookup falls back category → `other` → English → the key itself. `useT()` exposes `tp` next to `t`, keeping the works-without-a-provider behaviour.

| key | 1 | 2–4 | 5+ |
| --- | --- | --- | --- |
| `permission.meta.lines` | 1 строка | 3 строки | 5 строк |
| `permission.meta.chars` | 1 символ | 3 символа | 5 символов |
| `tasks.chip.running` | Выполняется 1 задача | Выполняются 2 задачи | Выполняется 5 задач |
| `tasks.chip.total` | 1 фоновая задача | 2 фоновые задачи | 5 фоновых задач |

The dictionary contract grows rather than loosens. `messagesParity.test.ts` derives plural families from the default locale (a base key spelled out as `.one` + `.other`) and requires every locale to carry exactly the categories its own `Intl.PluralRules` produces — deleting `permission.meta.lines.many` from `ru.ts` fails the suite with a `missing` entry. Token parity now walks each locale's own keys, so the extra Russian categories are checked against the family's default entry instead of being skipped.

English output is unchanged apart from `"1 chars"` becoming `"1 char"`. `AGENTS.md`, `DESIGN.md` and `docs/ui.md` document the plural contract alongside the existing parity rules.

## Verification

- `cd external/ui && node node_modules/vitest/vitest.mjs run` — 133 files, **744 tests passed** (739 before; +5 covering CLDR categories, English and Russian declension including the 11–14 exception and the 21 wrap-around, the unknown-family fallback, and the chip's counted label).
- `vite build` passed.
- `tsc --noEmit` reports 62 errors on this branch and 62 on `main` — no new type errors (the build does not run `tsc`).
- Prettier is clean on every file touched here.

PR #124, opened 2026-08-26
feat(ui,cli): live status line next to the typing dots

hijera:
## Why

While a turn runs, both surfaces only said "something is happening": the SPA showed three animated dots, the console a static `Working...`. Neither told the operator whether a file is being read, a command is running, the model is thinking, or nothing has come back for a minute.

Ported from [hijera/foxxy-agent](https://github.com/hijera/foxxy-agent), adapted to Coddy's tool registry.

## What the row shows

`Reading external/ui/src/ui/App.tsx · 12s` — verb, target, elapsed. States, in priority order:

| State | Phrase | Counter |
|---|---|---|
| Unresolved permission prompt | `Waiting for your approval` | no |
| Unresolved question prompt | `Waiting for your answer` | no |
| Running `tool_call` | verb + target (`Reading`, `Running`, `Searching`, …) | yes |
| In-progress `thinking` | `Thinking…` | yes |
| `memory_copilot` busy | `Working with memory` | yes |
| Waiting on the model | escalates with time (below) | yes |

A plain wait escalates: `Waiting for the model` → `The model is taking longer than usual` (15s, `typing-dots-status--slow`) → `Still no response from the server` (60s). The two prompt states carry **no counter** — nothing is running, so a climbing number would be a lie (same reasoning as `ToolCallMessage`'s frozen timer).

## SPA (`external/ui`)

- **`chat/liveStatus.ts`** — pure derivation from the transcript the SPA already holds, so **no backend, ACP or SSE change**. It scans back to the last `user_message` (a stale `in_progress` row from a finished turn must not drive the label forever after a branch switch or reload) and returns i18n *keys* plus the raw target, leaving translation and truncation to the component and keeping the tests locale-independent.
- **`toolCallTargetText`** in `permissionToolPreview.ts` names what a call acts on — `command` for `run_command`, `pattern` for `grep`, `path` for the fs family. It reuses the `normalizedToolName` / `parseArgsText` / `stringArg` helpers already in that file rather than adding a second parser.
- **`TypingDotsMessage.tsx`** — the counter ticks on a `setTimeout` aligned to the second boundary, then a `setInterval`, so the number never appears to skip; the step identity is stamped during render so the first frame counts from the right zero. The status node **must** stay after the three `.typing-dots-dot` spans, whose `:nth-child(2)/(3)` rules carry the bounce stagger. The component was hardcoded English and now goes through `useT()` like the rest of the SPA.
- **Layout** — only the target ellipsizes when space runs out; verb and counter are `flex: none`, so the row degrades rather than wrapping or clipping the verb.
- The phrase table is adapted to Coddy: no SVN or browser branches (foxxy has those tools, we don't), `config_*` added, prefixes `coddy_todo_` / `coddy_scheduler_` / `coddy_memory_`. Unknown ids — MCP tools included — fall back to `Running a tool` and keep the id as the target so the row stays debuggable.

## Console (`external/cli`)

- **`status.go`** is the Go twin of the phrase table. The two cannot share code across the language boundary, so both files carry a note that a tool added to one belongs in the other. The console has no i18n (every string in `header.go` / `footer.go` is a literal), so the phrases live there directly instead of behind keys.
- **`Loader.SetMessageFunc`** renders the message per frame, so the elapsed counter ticks off the loader's existing 80 ms tick rather than a second timer. `Render` runs on the UI goroutine (`RequestRender` → `renderCh` → `RenderThrottled`), so the provider reads `App` state with no extra synchronization.
- Transitions are driven from the same ACP updates the transcript already consumes: `ToolCallUpdate` / `ToolCallStatusUpdate` (verb, then target from the streamed arguments), reasoning chunks, `MemoryPhaseUpdate`, and the permission / question modals.
- One deliberate divergence: once assistant text streams, the SPA hides the dots entirely, but a console spinner has nowhere to hide — it says `Responding` instead.

## Scope

Two things from foxxy were left out on purpose:

- **`reconnecting` / `connectingMcp`** — these need an `mcp_phase` SSE event (`internal/acp`, `internal/session`, `external/httpserver`) and a reconnect-tracking module, neither of which exists in Coddy. Out of scope for this PR.
- **The `ui.status_line` toggle** — the row is always on. Adding the config field would mean a whole new `ui` section in `ui_schema.go` (there is none today) plus `docs/config.schema.json`, `docs/config-reference.md`, `config.example.yaml` and the `configure-coddy` skill.

## Tests

- **SPA: 780 tests / 134 files green.** New `chat/liveStatus.test.ts` (32 cases: verb mapping, path and command truncation, elapsed formatting, state priority, turn boundary, the 15s/60s thresholds); `TypingDotsMessage.test.tsx` extended with fake-timer counter tests, the slow modifier, the no-counter states, counter reset on step change, and the bare-dots fallback; `toolCallTargetText` cases added to `permissionToolPreview.test.ts`.
- **Console:** a happy-path scenario in `features/cli_tui.feature` drives `Waiting for the model` → `Reading README.md` → `Waiting for the model` through the real render, plus edge cases in `external/cli/status_test.go` per the repo's spec/unit split.
- `golangci-lint run ./...` and `--build-tags cli` — **0 issues**. Both binaries build (`cli` and `http,ui,cli,scheduler,memory`). `tsc --noEmit` reports the same 103 pre-existing errors as a clean tree — none added.

⚠️ `make test` is **not** fully green, but not because of this change: `TestConfigCommitAbortsWhenStagingCannotBeConsumed` and `TestConfigCommitKeepsStagingConsumedWhenRollbackFails` in `internal/tools` fail on Windows on an untouched tree too (verified by stashing). Every other tag combination passes.

## Screenshots

**Not attached — the browser pane in the authoring environment could not composite frames, so `screenshot` timed out on every attempt.** In their place I ran the geometry check from `.claude/rules/ui-verification.md` step 3 against a live `vite` server with the real `styles.css`, at **1280px** and **390px**, in **light** and **dark**:

- no horizontal overflow at either width, and `document.scrollWidth === clientWidth`;
- the long path ellipsizes at 390px (`scrollWidth > clientWidth` on `.typing-dots-status-target`) while **no** verb is clipped at either width — the intended degradation;
- the `@media (max-width: 480px)` rule applies (13px → 12px);
- both themes resolve their `color-mix(… var(--coddy-blend-base))` tokens to real colors with the intended hierarchy (target darker than the base status color, slow verb darker still, timer lighter), and the `\00b7` separator renders.

Please add images before merge if the repo gate requires them.


EvilFreelancer:
Pushed three follow-up commits (65bcba5..87e33d4) fixing the defects a cross-agent review of this PR converged on (Claude, Codex CLI, and Cursor independently flagged the same console findings). Each fix landed test-first: the reproduction fails on the previous code.

## b1bddaf — fix(cli): keep the status line truthful through permission and question gates

The modal phrase was an ordinary `setStatus` call into a last-writer-wins field, which broke three ways against the real event order from `internal/agent/react.go` (`in_progress` is emitted **before** `RequestPermission`, and nothing else is emitted until the tool finishes):

- `updatesCh` (buffered 1024) races `permCh` (unbuffered) in the UI `select`: when the modal wins the select, the queued `in_progress` update overwrites "Waiting for your approval" with a ticking `Running <cmd>` while the modal is still open;
- in the other ordering the blocked phrase survives, but after Allow `closeModal` reset to "Waiting for the model" although the approved tool only starts executing at that moment — a long command escalated to the false "Still no response from the server";
- the `question` tool's verb is the same string as its blocked phrase, so the `setStatus` dedup kept `counts=true` and the counter ticked while the operator typed an answer.

The blocked phrase is now an overlay (`blockStatus` / `unblockStatus` in `external/cli/status.go`) kept apart from `stepStatus`: it always wins while a modal is open, cannot be overwritten by a late update, and renders with no counter. Lifting the gate returns the line to the gated step and restarts its clock, so operator thinking time is never billed to the tool.

Observed live on a real gated `sleep 6` run (screen hardcopy sampled every ~0.7 s), **before**:

```
(permission modal open)    Running sleep 6 · 0s → 1m 16s     counter ticks under the open modal
(after Allow, executing)   Running sleep 6 · 1m 17s → 1m 22s inflated clock during real execution
(completed)                Waiting for the model · 0s
```

**after**:

```
(permission modal open)    Waiting for your approval          whole period, no counter
(after Allow, executing)   Running sleep 6 · 0s → 5s          clock restarts when the gate lifts
(completed)                Waiting for the model · 0s → Responding
```

Coverage: a new scenario in `features/cli_tui.feature` ("The status line stays truthful through a permission gate") drives pending → in_progress → permission → Allow → completion through the real render and fails on the previous code exactly on the post-approval assertion; unit tests cover the overlay-vs-late-update race, the question verb collision, and the no-gate `closeModal` path (model picker and friends must not touch the step clock).

## 9ed2802 — fix(ui): prefer the running tool over a later pending one

`deriveLiveStatus` took the first pending-or-in_progress hit while scanning backward, so with the agent's announce-all-pending-then-execute-sequentially pattern the newest `pending` row masked the call actually running. `in_progress` now wins; a `pending` call drives the label only when nothing runs yet. Regression test added.

## 87e33d4 — docs

`docs/cli.md` still described the static `Working...` spinner; `docs/ui.md` had no functional entry for the row at all. Both now match the shipped behavior (no-counter modal states, 15 s / 60 s waiting escalation, pointers to the twin phrase tables).

## Verification

- full `make test` matrix green on Linux; the two `internal/tools` failures mentioned in the PR description are Windows-only and reproduce on an untouched tree, as stated;
- vitest: 781 tests / 134 files green;
- `make lint` 0 issues on both passes (including `--build-tags cli`);
- SPA behavior verified against a live backend via Playwright: waiting → thinking → permission (no counter) → running with a restarted counter → the 15 s / 60 s waiting escalation, all as specified, in both themes and at 390 px / 1280 px.

Deliberately **not** bundled here (minor polish, separate from the defects): the brief SPA counter on the `question` tool after its prompt resolves, `waitingFrom` ignoring completed assistant messages, the 56 vs 48 target-length drift between the TS and Go tables, and the elapsed counter being `aria-hidden`.

PR #140, opened 2026-09-04
feat(composer): line-range @mentions with a picker for manual entry

hijera:
## What

An `@` mention can now narrow a workspace file to a **1-based inclusive line range** — `@Dockerfile:21-31`. The composer chips the whole token, and the model receives only those lines as `<coddy_attachment path="Dockerfile" name="Dockerfile" lines="21-31">`. On the Dockerfile in this repo the same mention went from attaching **2424 chars / 76 lines** to **332 / 8**.

Typing the `:` opens a **line-range picker**: it previews the file and highlights as the digits are typed. On desktop the rows are clickable and a drag writes the range back into the text; on mobile shells they are display-only and the range is typed.

## Why this shape

Ported from [hijera/foxxy-agent#54](https://github.com/hijera/foxxy-agent/pull/54), **minus its IDE half**. That PR turns a paste into a chip by asking editor plugins what was just copied (an `internal/idecopy` ring plus `/ide/copy-buffer` and `/ide/paste-classify`). This repo has no plugins to ask — no `editors/`, no `internal/ideenv`, no IDE routes — and the OS clipboard cannot stand in for them:

- The clipboard stores formats, not provenance. Copying text out of an editor puts `CF_UNICODETEXT` / `text/plain` (often `text/html`, RTF) on it; there is no "which file did this come from" format on Windows or Linux.
- `CF_HDROP` / `text/uri-list` do carry paths, but only when you copy **files in a file manager**, not text from inside one.
- Private IDE formats do not help either: JetBrains puts JVM-local flavors on it, VS Code/Electron writes `text/html` with no file path.
- And a web page could not read them regardless — `ClipboardEvent.clipboardData` exposes a fixed allowlist of MIME types.

So the token grammar and the line slicing port cleanly; the entry point is manual instead. `@terminal` from that PR is **not** ported — this repo has no terminal mentions, it would be dead code.

## How

- **Grammar**, two twins sharing their test literals: `external/ui/src/ui/skills/draftAt.ts` absorbs a `":N-M"` suffix into the `@path` span (so `segmentComposerMirrorSpans` chips the whole token with no change of its own), and `internal/session/at_paths_extract.go` gains `ExtractAtFileRefsFromText` with the same rules. A range must end the token (`:21-31x` stays prose) and satisfy `1 <= start <= end`; `@foo :1-5` does not attach, since that token had its trailing space trimmed. `ExtractAtFilePathsFromText` keeps its old contract for `hydrate_plans.go`.
- **Hydration**: `PromptFileAttachmentSourceField` gains `StartLine`/`EndLine`; the range rides `acp.Resource.URI` as a `#L21-31` fragment (`lineRangeURI` / `splitLineRangeURI` / `sliceLines`) so the ACP types are untouched, and `resourceBlockToXMLAttachment` renders it as the `lines` attribute. Byte offsets (`Start`/`End`) keep winning where both are present. A plain mention and a ranged one of the same path are distinct attachments.
- **SPA**: `App.tsx` sends `attachments[].source.startLine` / `endLine`; the backend always reads the range from the file (no paste literal to carry). `stripCoddyAttachments.ts` collapses a ranged block back to `@path:N-M`.
- **Picker**: `:` already closes the file picker (`:` is no `MENU_PATH_CHAR`), so the panel takes its place through the same portal / `slash-menu--sheet` chrome. Pure logic lives in `external/ui/src/ui/skills/draftAtRange.ts` (`atRangeDraftAtCaret` / `replaceAtRangeSuffix` / `highlightedRange`); the composer text stays the only input, so there are no number fields to keep in sync. A path that does not resolve leaves the panel closed, so `@user:1-2` in prose never pops an empty panel, and the settled path is remembered so the next digit refetches nothing. `Escape` dismisses and suppresses it for that mention; `Enter` is left alone and still sends.
- **New route**: `GET /coddy/workspace/file?path_rel=&max_lines=` over `session.ReadWorkspaceUTF8`, which already owns path normalization, the traversal guard, the 512 KiB cap and legacy-encoding decoding. `openapi.go`, `docs/http-api.md` and `docs/ui.md` updated.
- **Bonus fix**: derived session titles no longer leak `<coddy_attachment>` bodies (pre-existing, but ranged mentions make it louder).

## Tests

- BDD: `features/at_line_range_mention.feature` with a godog harness in `internal/agent` that drives both hydration entry points and renders the blocks the way a turn does, so the assertions read the XML the model actually receives.
- New/extended unit tests: `at_line_range_test.go`, `promptfiles_lines_test.go`, `filesystem_test.go`, `react_test.go`, `workspace_file_test.go` (http tag), and vitest for `draftAt`, `draftAtRange`, `composerMirrorSegments`, `stripCoddyAttachments`, plus six `Composer.test.tsx` cases covering the panel (opens on `:`, highlight follows the digits, desktop drag writes the range, backwards drag normalizes, mobile rows are not buttons, unresolvable path stays closed).
- `make test` green across the full tag matrix, `make lint` clean, SPA suite 937/937.
  - **Two pre-existing Windows failures**, unrelated to this change: `TestConfigCommitAbortsWhenStagingCannotBeConsumed` and `TestConfigCommitKeepsStagingConsumedWhenRollbackFails` in `internal/tools` rely on POSIX directory permissions to block a delete, which Windows does not honour. Verified identical on a clean worktree at `main`.
- Live E2E on a running `http,ui` build: typed `@Dockerfile:5-12` → panel opened and highlighted; dragged rows 2–4 → composer became `@Dockerfile:2-4`; sent → the persisted turn carried `lines="2-4"` with exactly those lines, and the bubble showed `@Dockerfile:2-4`.
- Layout per `.claude/rules/ui-verification.md`: at 1280 the panel edges match `.composer-card` within 1px, and `scrollWidth - clientWidth == 0` at both 390 and 1280 — long code lines scroll inside their own container, never the page. `/layout-scroll-check.html` is **not in the repo** (404 on the running build), so only `/` was checked.

## Screenshots

Twenty PNGs were captured from the running `http,ui` build — 390 and 1280, light and dark, with before/after pairs against `main` for the two surfaces that already existed:

- `range-picker-{wide,narrow}-{dark,light}` — the new panel, anchored on desktop and as a bottom sheet at 390 (rows display-only there).
- `before-range-mention-*` vs the panel shots — on `main` the same token chips only `@Dockerfile` and leaves `:5-12` as prose.
- `at-menu-*` vs `before-at-menu-*` — the `@` picker gains one footer hint line, `«:» после файла — выбор строк`.
- `before/after-user-bubble-wide-{dark,light}` — visually identical here by design (the strip helper suppresses the injected `@path` when the text already mentions it); the difference is in what was attached, measured above.

**The GitHub CLI cannot attach images to a PR body.** The files were delivered in the working session and can be dropped into this description from there; the visual delta is described per pair above so the change is reviewable without them.


EvilFreelancer:
## Review round: merge with `main`, three-way cross-review, follow-up commits

`main` is merged into the branch (conflicts were only the two test files where both sides appended tests). The PR was then reviewed independently by Codex (gpt-5.6-sol), Cursor (auto) and coddy (neuraldeep/qwen3.8-27b) against the same brief with the full diff, and every finding was re-checked by hand before acting on it. The premise holds for all three reviewers: whole-file mentions are the wrong granularity for long files, the OS clipboard cannot supply file provenance, and `@path:N-M` plus a picker is the right entry point for this repo.

### Findings that were real and are fixed in the follow-up commits

- **A range the file cannot honour widened into the whole file with a wrong label.** `@f.txt:99-100` on a 5-line file attached lines 1-5 under `lines="99-100"`. Now such a range (zero, inverted, or starting past the last line) is `ErrLineRange`: `POST /v1/responses` answers `400`, an ACP `resource` with such a `#L` fragment fails the prompt like a missing file, and a range typed into prose stays prose (nothing is attached, the model can still read the file). An end past the last line still clamps.
- **The `lines` label described bodies that were not those lines.** A `source.literal` or byte-offset body carried `#L…` too. The label is now written only when the body really is the slice.
- **Two parsers of the `#L` fragment disagreed.** `fmt.Sscanf` accepted `notes#L1-2.go` and `f.go#L1-2junk`, while the XML renderer used an anchored regex and would emit `lines="9-2"`. One strict, exported `session.SplitLineRangeURI` (whole tail only, `1 <= start <= end`) now serves both hydration and `resourceBlockToXMLAttachment`; a file literally named `f.go#L5` or `notes#L1-2.go` resolves as a file.
- **Derived titles could leak an attachment body.** The regex stripper stopped at a `</coddy_attachment>` inside CDATA. Replaced with a scanner that skips CDATA sections (the renderer splits `]]>` across two of them), with a regression test.
- **The picker kept another session's preview.** The loaded-path cache was keyed by path only; a `sessionId` change now closes the panel and discards in-flight reads (vitest case added). The preview is documented as a snapshot; the attached lines are read at send time.
- **`stripCoddyAttachmentsForUserDisplay` turned `lines="9-2"` into `@f.go:9-2`.** Malformed labels fall back to the plain `@path`.
- **`GET /coddy/workspace/file` answered `400` for unexpected I/O errors** while the spec says `500`. Classified now, with a test.
- **Flow gaps:** `DESIGN.md` has the line-range picker contract; `docs/ui.md` named the wrong extractor; `docs/cli.md` notes that the console hydrates `@f.go:2-4` the same way (the picker is SPA-only); the two ad-hoc `internal/session` test files are folded into `at_paths_extract_test.go`, `promptfiles_test.go` and `filesystem_test.go`; the Go dedupe key matches the URI shape.
- **Real-binary harnesses** `examples/httpserver/http_e2e_mentions.py`, `examples/acp/acp_e2e_mentions.py`, `examples/cli/cli_e2e_mentions.py` (wired into the three runners and `examples/README.md`): picker read, `attachments[].source.startLine/endLine`, the typed grammar, an editor-style `resource` with a `#L5-5` fragment, and the `400` / prompt error for a range past the end. All three pass against a live model.

### Findings checked and not acted on

- Cursor: "TS `\d` with the `u` flag accepts Unicode digits, so the twins diverge." False: JavaScript `\d` is ASCII-only regardless of `u` (`/^:(\d{1,9})-(\d{1,9})/u.exec(":٢-٤")` is `null`), so both grammars accept the same digits.
- Codex: carry the range structurally instead of a URI fragment. The ACP resource type is the protocol's, and the strict tail-only parser removes the practical ambiguity; a file whose name ends in a well-formed `#L<n>-<m>` remains the one documented edge.
- Codex: Windows-only dedupe divergence for `@dir\f.go @dir/f.go` (`filepath.ToSlash` on the Go side). Left as is: the SPA cannot know the server platform, and the attachment list itself is unaffected.
- Symlink containment in `ReadWorkspaceUTF8` is lexical. Pre-existing and shared with every workspace read, so out of scope here.

### Verification

`make test` (full tag matrix), `make lint`, vitest 145 files / 939 tests, and the three new e2e harnesses are green. One matrix run tripped on `TestSpawnSubagentChildSpawnNeverNotifies` under `-tags=http,cli`, a known flake untouched by this PR (15 consecutive runs green, the full matrix green on rerun). The Windows CI job failed in the console TUI scenario `Starting a new session drops updates from the old one` on the fake terminal; `external/cli` is not part of this PR and the same job fails intermittently on other branches, so it was re-run. Layout per `.claude/rules/ui-verification.md` on the running `http,ui` build: at 1280 the panel edges sit 1px inside `.composer-card`, at 390 it is the full-width sheet with display-only rows, and `scrollWidth - clientWidth` is 0 on the page at both widths (long lines scroll inside `.at-range-lines`). Screenshots (390/1280, light/dark, the `@` menu hint) were captured from that build and delivered in the working session, since `gh` cannot attach images to a PR.

PR #177, opened 2026-09-10
Scope log levels per component and make the Telegram command path readable

EvilFreelancer:
## Why

An operator chasing a Telegram command that does not work has nothing to read. `logger.level: debug` barely differs from `info` because the whole repository had 19 `.Debug()` calls, 13 of them in the gateway and none on the command or callback path: `/model` builds a keyboard, the tap is acknowledged, and the log stays silent whether it worked or not.

Two defects behind "the model does not change" are fixed here as well, each pinned by its own scenario.

## What

**`logger.levels` — per-component minimum severity.** A list of `{component, level}` entries, matching how the repository already shapes name-value pairs (`mcp_servers[].env`, `headers`), which also means the Settings form renders it from the shared schema with no new component:

```yaml
logger:
  level: "info"
  levels:
    - component: "gateway.telegram"
      level: "debug"
```

A component is the dotted name a subsystem tags its logger with. A parent covers everything nested under it and the longest configured prefix wins, so `gateway` reaches `gateway.telegram` unless that name carries its own entry; records from untagged code follow `level`. Configuring one component twice is an error rather than a silent winner.

`--log-level` takes the same spec as a comma-separated list, which is how an operator under systemd raises one subsystem for a single restart:

```bash
coddy gateway --log-level "info,gateway.telegram=debug"
```

A bare level leaves the configured entries alone; a spec that names components replaces them.

**How it works.** `slog` decides whether to build a record from `Enabled(ctx, level)`, which never sees attributes, so the format handler is opened at the lowest level any component asks for and `componentHandler` does the real filtering. The component of a handler instance is fixed by the `WithAttrs` call that tagged it, so resolution happens once per logger, not once per record — and with no overrides configured the wrapper is not inserted at all. Components are tagged once, at construction: `gateway` / `gateway.telegram` in `external/gateway/start.go`, `session` in `NewManager`, `agent` in `NewAgent`, `scheduler` in the daemon (which previously stamped `component=scheduler` inline, which reads the same in a file but cannot scope anything).

**The Telegram debug trail.** Every update as it arrives, a `reason` for every silent drop (access denied, admin-only chat, a group message not addressed to the bot, a full worker queue), each recognised command, each menu with the session it belongs to, and each callback with the model or mode it resolved to. A switch that lands is `info` (`telegram: model applied`, matching `telegram: session cleared`), so the confirmation is visible without raising anything; a failure is `warn`. `reply` no longer logs through `slog.Default`, where a failed send missed the configured sink entirely.

**Two `/model` defects.**

- `handleCallback` took the session id from the store but never loaded the session. The keyboard outlives the process that sent it, so a tap after a restart addressed a session that is on disk but not live, and the manager only configures live ones. It now calls `ensureSession` first.
- `callback_data` was truncated to 64 bytes, sending back a model name nothing is configured under. A long id now travels as a digest and is resolved back against the configured models; a model dropped from the config since the keyboard was sent is reported rather than applied.

`Store.Peek` was added so the new log line reports an existing mapping without minting one for a chat that only typed `/help`.

## Tests

- `features/logger_component_levels.feature` — raising and lowering one component, parent coverage, longest-prefix precedence, untagged records, the `--log-level` spec.
- `features/gateway_telegram_model_switch.feature` — the `/model` happy path, a button tapped after a restart, a model id over the callback limit, and the debug trail itself. Both defects were verified to be pinned: reverting either fix fails exactly its scenario.
- Unit tests for the edges: bad and duplicate `levels` entries, spec parsing, `ApplyOverrides` merge rules, the handler's `Handle`/`WithGroup`/`minLevel` behaviour, callback payload round-trips at and over the limit, `Peek` not minting, and that each constructor really tags its logger.

## Docs

`docs/config-reference.md` (field table plus the `levels` and `--log-level` narrative), `docs/config.md`, `docs/gateway.md` (new **Debugging a chat** section), `docs/cli.md`, `config.example.yaml`, `docs/config.schema.json`, `UISchemaMap`, the bundled `configure-coddy` skill, the zsh completion, and `.claude/rules/gateway.md` with its `.cursor` mirror.

The verbatim schema mirror in `coddy-project.github.io` needs the same `logger.levels` block; it is prepared on a local branch there and is not part of this PR.

## Screenshots

Settings → System → Logger, the new **Component levels** editor. It comes from the shared array editor, so no SPA source changed — the surface is new because the schema is.


EvilFreelancer:
Screenshots of the new **Component levels** editor (Settings → System → Logger), captured against a live `coddy http` with the SPA on `vite`:

- `settings-logger-levels-wide.png` — the field with one configured entry, 1280px, dark
- `settings-logger-levels-add.png` — a row added with **Add**: it seeds at `debug`, leaving only the component name to fill in
- `settings-logger-levels-narrow.png` — 390px, controls stack with no horizontal overflow
- `settings-logger-levels-light.png` — light theme

No SPA source changed: the surface is new because `UISchemaMap` describes the field and the shared array editor renders it, so there is no before image of the field itself.

The images cannot be uploaded through the GitHub API, only by dragging them into the web editor. They are in the branch worktree under `.playwright-mcp/`.

Also verified live rather than only in tests: the round trip SPA → `PUT /coddy/config` → `config.yaml` keeps the entries, and an entry with an empty component is rejected by `POST /coddy/config/validate` with `logger.levels[2].component: required (for example "gateway.telegram")`.

PR #213, opened 2026-09-12
fix(cli): operator prompts queue for the modal slot instead of replacing each other

EvilFreelancer:
## The report

A user in the console: generation suddenly stopped, the token counters froze, no
question form ever appeared. He opened the web UI and found the same session
sitting on an unanswered question.

## What actually happens

The console has one modal slot (`App.modal`) and every opener took it
unconditionally. When a second operator prompt arrived while the question tool
held the slot, `openModal` replaced the question modal outright:

1. the question modal opens, the turn worker blocks on its reply channel;
2. a second gate arrives and takes the slot; the question modal leaves the tree
   and its reply channel is orphaned;
3. answering the second one calls `closeModal`, so the editor comes back and the
   screen looks idle;
4. the question's worker is still blocked on `<-req.reply`. Nothing on screen can
   send it. The spinner runs, no updates arrive, and the `question` tool call
   stays `in_progress` in the persisted transcript — which is exactly what the
   web UI renders as "a question is waiting".

The path that reaches it in practice is a `spawn_agent` child started with
`background: true`: the parent does not wait for it, so the parent can be parked
in the question tool while the detached child relays its own permission request
to the same surface.

## The fix

Prompts somebody is blocked on (permission requests and questions) queue for the
slot in arrival order:

- the prompt on screen keeps it and names how many are waiting behind it;
- each queued prompt is shown as the previous one is answered;
- a prompt whose turn ended while it waited is dropped instead of shown to nobody;
- selectors go through `openOverlay` and never take the slot from a prompt (a
  dropped picker costs a keystroke; a dropped prompt hangs a turn). The `/resume`
  list is fetched asynchronously, so when it is refused the console says so
  instead of looking like the command did nothing;
- a turn ending closes only a modal whose own requester is gone, so a background
  child's prompt outlives the turn that spawned it.

## What the operator sees

The modal, rendered by the real component at 96 columns. A single prompt is
unchanged, to the row:

```
────────────────────────────────────────────────────────────────────────────────
 Where to put the relay
 The swarm relay has to live somewhere the nodes can reach. Which placement do you want?
→ Relay and node both on nas02 (recommended)
  Relay on a public VPS, nodes dial out

  One box carries the relay and the first node; simplest to operate.
 ↑↓ navigate · enter choose · esc cancel
────────────────────────────────────────────────────────────────────────────────
```

With a background subagent's permission queued behind it:

```
────────────────────────────────────────────────────────────────────────────────
 Where to put the relay
 The swarm relay has to live somewhere the nodes can reach. Which placement do you want?
→ Relay and node both on nas02 (recommended)
  Relay on a public VPS, nodes dial out

  One box carries the relay and the first node; simplest to operate.
 1 more prompt is waiting behind this one
 ↑↓ navigate · enter choose · esc cancel
────────────────────────────────────────────────────────────────────────────────
```

Answering the question promotes the one that was waiting:

```
────────────────────────────────────────────────────────────────────────────────
 Permission required
 [subagent explore] Run: rm -rf /tmp/scratch

→ Allow
  Reject
 ↑↓ navigate · enter choose · esc reject
────────────────────────────────────────────────────────────────────────────────
```

## Tests

Happy path in `features/cli_tui.feature`, driven through the real app loop with a
background subagent asking permission while the question is open. Without the fix
it fails on the symptom:

```
Then the question is still the prompt on screen
  Error: modal = *cli.permissionModal, want the question still holding the slot
```

Edge cases as unit tests in `external/cli/chat_test.go`: two prompts queue and both
get answered; a queued prompt whose turn ended is dropped; a picker cannot take the
slot from a prompt; a background child's prompt survives its spawning turn ending.

`make test` green (49 packages, no failures), `make lint` clean, `make docs-check`
clean.

## Notes

- The console screenshot set on `docs/surfaces/console.md` covers the launch
  frame, the model selector and the usage footer; it has never shown the question
  or permission modal, so this change updates the prose of that page and does not
  add a one-off modal capture. The rendered frames above are the evidence.
- `make site-docs-check` reports `llms.txt` / `llms-full.txt` drift, but it
  reports the same on `main` with this branch's docs stashed, so the site layer
  was already behind before this change. Per the workflow the site commit follows
  the merge anyway.
- Not addressed here, because it is a product call rather than a defect: the
  `question` tool still blocks a turn under `permission_mode: bypass`. The user
  expected bypass to cover questions too. Say the word and I will make bypass
  suppress or auto-answer them.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -3523,6 +3523,14 @@ a column boundary. When the question is answered, its tool block shows the
 questions with the chosen answers (`→ answer`, `→ (no answer)` for a dismissed
 one) rather than the JSON the tool hands the model.
 
+A permission request and a question are prompts a turn is blocked on, so they
+queue for the modal slot rather than replace each other: the one on screen
+keeps it and says how many are waiting behind it, and each is shown in arrival
+order as the previous is answered. A selector never takes the slot from either
+of them - losing a picker costs a keystroke, losing a prompt would leave its
+turn waiting for an answer with nothing on screen to give it. A prompt whose
+turn ended while it waited is dropped instead of shown.
+
 ## Flags
 
 `--config --home --cwd --sessions-dir` mirror the other subcommands.
@@ -3634,9 +3642,11 @@ process working directory, resolved like `coddy mcp`. Under
 
 In the console a `spawn_agent` call shows as a tool box like any other, and the
 status line reads `Running subagent <name>` with its elapsed counter for as long
-as the child runs. A child's permission request, while its spawning turn is
-still alive, opens the usual modal in the parent chat with the title prefixed
-`[subagent <name>]`. Child sessions (`sub_…` ids) are read-only transcripts:
+as the child runs. A child's permission request opens the usual modal in the
+parent chat with the title prefixed `[subagent <name>]`, waiting its turn when
+the parent is itself holding a prompt. A child started in the background
+outlives its spawning turn, so its request stays on screen after that turn
+ends. Child sessions (`sub_…` ids) are read-only transcripts:
 `-c` never picks one, and a prompt sent to one is refused with a message naming
 the parent session.
 
--- a/docs/surfaces/console.md
+++ b/docs/surfaces/console.md
@@ -217,6 +217,14 @@ a column boundary. When the question is answered, its tool block shows the
 questions with the chosen answers (`→ answer`, `→ (no answer)` for a dismissed
 one) rather than the JSON the tool hands the model.
 
+A permission request and a question are prompts a turn is blocked on, so they
+queue for the modal slot rather than replace each other: the one on screen
+keeps it and says how many are waiting behind it, and each is shown in arrival
+order as the previous is answered. A selector never takes the slot from either
+of them - losing a picker costs a keystroke, losing a prompt would leave its
+turn waiting for an answer with nothing on screen to give it. A prompt whose
+turn ended while it waited is dropped instead of shown.
+
 ## Flags
 
 `--config --home --cwd --sessions-dir` mirror the other subcommands.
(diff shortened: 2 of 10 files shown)


EvilFreelancer:
CI: 28/28 green after a rerun of one job.

The first run failed `Test suite (gateway)` on `TestTaskOutputAndMetadataPersistUnderTheSessionDir` in `internal/bgtask` — a package this branch does not touch (the whole diff is `external/cli` behind the `cli` tag, which the `gateway`-only build does not compile).

It is a pre-existing race in the test, not a regression. The supervisor sets the final status on `t.snap` under `t.mu`, unlocks, and only then calls `p.persist(t)`. The "Persist before releasing waiters" ordering in `pool.go` protects callers that block on `Wait`/`Stop`, because those wake after the persist; it does not protect a caller that *polls* the status. `waitForStatus` polls, so between the unlock and the persist it can see `succeeded` in memory while the metadata file still holds the initial `running` record, and `LoadPersisted` rewrites any unfinished record as `orphaned`. That is exactly what the log showed (`Status:orphaned` with `FinishedAt:<nil>`, `ExitCode:<nil>`).

Not reproducible locally: 40 plain runs and 60 under `-race` on that one test all pass, so it needs a loaded runner. Filed separately rather than fixed here, to keep this PR to the console change.

EvilFreelancer:
Closing without merge. The deadlock this fixes is real, but the fix keeps the console as the owner of its prompts, which is the model we are moving away from: a permission request or a question is going to be session state on a session bus in the manager, visible to every attached surface and answerable from any of them. The console-side queue logic here will be recast as the console consumer of that bus. Design record and work plan follow in a separate PR (docs/plans/session-bus.md).

PR #219, opened 2026-09-12
feat(prompts): model-tuned system prompts, Gemma 4 on NeuralDeep

hijera:
## What

Ports the model-tuned system prompts of [hijera/foxxy-agent](https://github.com/hijera/foxxy-agent) (`internal/prompts` at `origin/main` 0e27bfe2) and tunes the Gemma variant against Gemma 4 served by NeuralDeep (`gemma-4-31b`, `gemma-4-31b-noreason`). Three commits:

1. **`refactor(prompts)`: section assembly.** `agent.md`, `plan.md` and `ask.md` become ordered fragments under `internal/prompts/sections/<mode>/`, listed by a `manifest` per mode (`internal/prompts/sections.go`). The text does not change: the built-in render of agent, plan, ask and an unknown mode was compared with the monolith render for four data sets (empty, typical, all fields, subagent role) and is identical once runs of blank lines are collapsed. The conditional blocks use `{{-`, so empty blocks no longer leave runs of blank lines.
2. **`feat(prompts)`: model-tuned prompts.**
   - **Variant keys.** Before every model call the agent resolves variant keys, most specific first (`Agent.promptVariants`): the model reference slug (`neuraldeep-gemma-4-31b`), the API model slug (`gemma-4-31b`) and the family from `prompts.Family`. The model id wins over the provider type, because open models are served through OpenAI-compatible hubs.
   - **What a variant can do.** It fills the optional `notes` / `model_notes` slots, replaces a fragment (`<id>_<variant>.md`) or brings its own `manifest.<variant>`. Under `prompts.dir` the keys select `<mode>.<key>.md` before the base file, and the rules/instructions dedupe follows the file that is actually rendered.
   - **Shipped variants**, ported from foxxy and adapted to Coddy's three modes and tool names:
     - agent-mode family notes for `anthropic`, `openai`, `gemini`, `gpt-oss`, `qwen`, `neuraldeep` and `gemma`;
     - the `openai` plan header, tool list and planning steps;
     - the `openai` and `gpt-oss` ask structures;
     - `gpt-oss-20b` / `gpt-oss-120b` profiles in every mode;
     - the Gemma rules described below.
   - **Config switch.** `prompts.per_provider.enable` (default on; `false` sends every model the shared prompt) is carried through the struct, `config.schema.json`, the JSON DTO, the UI schema with en/ru strings, `config.example.yaml`, the configure-coddy skill and the docs (`docs/features/modes.md` → *Model-tuned system prompts*).
   - Section fragments are read with LF line endings whatever `core.autocrlf` did to the checkout.
3. **`fix(prompts)`: real tool names.** The shared agent sections named `apply_diff`, `search_web` and `extract_page_content`, none of which is a registered tool; they now name `edit` / `apply_patch`, `websearch` and `webfetch`. `TestBuiltInPromptsNameOnlyRegisteredTools` pins every mode against the registry.

## Gemma 4 on NeuralDeep

The measurement comes from the new live probe `internal/agent/live_tool_json_test.go`. It is skipped unless `CODDY_LIVE_MODEL` and `NEURALDEEP_API_KEY` are set; `CODDY_LIVE_PROMPTS=shared,tuned` compares the two prompts. It runs five coding tasks whose arguments are easy to break as JSON:

- an edit with regex backticks, `\"`, `\\` and `\t`;
- a nested JSON document written verbatim;
- a Russian-language checklist with integer indexes;
- paging with integer `offset`/`limit` plus `keep_result`;
- a two-file rename with `replaceAll`.

For every run it counts invalid JSON arguments, schema violations against the real tool schemas (types, required, enums, unknown properties), tool errors, calls that ended up as text, and empty turns, and checks the result on disk.

**What breaks.** Gemma 4 produced valid, schema-conforming JSON in every structured call. The failure is elsewhere: **the hub turns only the first tool call of a message into a structured call and hands a second one back as text**, for example

```
<|tool_call>call:coddy_todo_item_update{index:1,status:<|"|>completed<|"|>}<tool_call|>
```

That text lands in the answer and the call never runs, so the checklist or the rename is left half done.

**Iterations of the Gemma notes.** The first version (one call per message) removed the lost calls, but a message that announced its next call in words sometimes ended without the call: 2 of 3 renames failed that way. The final version:

- asks for one bare tool call per message and no text around it;
- never writes a call as text;
- fills arguments exactly as the schema defines them, with no escaping of its own (the native call format has its own string delimiters, and the hub does the JSON encoding);
- replaces the shared "explain your reasoning before each tool call" rule in the agent's *How to work*.

**`gemma-4-31b-noreason`**

| Prompt | Runs | Invalid JSON | Schema violations | Tool errors | Calls lost as text | Empty turns | Tasks done |
|---|---|---|---|---|---|---|---|
| Before the port (monolith prompt) | 10 | 0 | 0 | 0 | 4 | 0 | 9/10 |
| Shared prompt after the port (`per_provider.enable: false`) | 15 | 0 | 0 | 0 | 4 | 5 | 14/15 |
| Gemma prompt | 21 | 0 | 0 | 0 | 0 | 0 | 21/21 |

**`gemma-4-31b`**

| Prompt | Runs | Invalid JSON | Schema violations | Tool errors | Calls lost as text | Empty turns | Tasks done |
|---|---|---|---|---|---|---|---|
| Before the port (monolith prompt) | 10 | 0 | 0 | 0 | 3 | 2 | 10/10 |
| Gemma prompt | 16 | 0 | 0 | 0 | 0 | 0 | 16/16 |

**Hub stalls, left out of the tables.** Through the session NeuralDeep repeatedly sent nothing within the 90 s first-token limit for some request bodies; Coddy re-issued each one once and then gave up with `model did not respond`. Those runs made no tool calls and are not in the tables. That is also why there is no post-port shared-prompt row for `gemma-4-31b`: its runs stalled. The same requests, dumped from the probe and replayed outside Coddy, answered within seconds, so I read the stalls as hub-side rather than caused by the prompt text.

## Verification

- **`make test` before the change** (`ceda32ab`, this Windows machine): 43 packages ok. 6 packages fail with 8 Windows-only tests:
  - `TestNeuralDeepLoginFlowFeature`, `TestFileSecretStoreRoundTrips`: file mode 0600;
  - `TestPackagingFilesTrackTheCommandSet`: CRLF man page;
  - `TestLoadLegacyLoggerFileAddsOutputs`, `TestPrepareExportOutput`: path separators;
  - `TestExplicitSkillsDirMissingIsAWarningDefaultsAreSilent`: a backslash path in YAML;
  - `TestConfigCommitAbortsWhenStagingCannotBeConsumed`, `TestConfigCommitKeepsStagingConsumedWhenRollbackFails`: the config commit rollback paths.
- **`make test` after the change** (final tree): the same 43 packages ok and exactly the same 8 failures, nothing new. The Linux CI matrix is the authority here.
- **Unit tests** (edge cases): `internal/prompts` (`family_test.go`, `slug_test.go`, `variants_test.go`: variant precedence in `prompts.dir`, the configured file name, fallbacks, manifest overrides, optional slots, structural pins, every family × mode keeping the identity line once inside the gateway window, the subagent role and no blank-line runs, per-model profiles, the Gemma rules, no foxxy wording); `internal/agent/system_prompt_variants_test.go` (`promptVariants`, the switch, per-model files, the rules/instructions dedupe with a variant file); `internal/config/prompts_test.go` (default, explicit false, YAML, JSON DTO round trip); offline checks of the probe's validator and text-call patterns.
- **BDD** (happy path): `features/model_tuned_prompts.feature`, 4 scenarios.
- **`make lint`**: 0 issues. **UI dictionaries**: `npx vitest run src/ui/i18n`, 34 passed (parity included). `git grep -n -i 'агентск'`: only the wording rule itself.
- **Docs**: config tables and `llms-full.txt` regenerated with `docsgen -write -skip-cli`. The CLI reference is untouched. On this checkout the docsgen link check reports `docs/contributing/custom-tools.md: broken link argsJSON` before and after the change.

## Screenshots

Settings → System → Prompts gains the *Model-tuned prompts* group (Dark, 1280).

| Before | After |
|---|---|
| before, English | after, English |
| before, Russian | after, Russian |

## Notes

- The schema only adds an optional key. Publishing it and the docs layer to the site (`make site-schema`, `make site-docs`) needs the `coddy-project.github.io` checkout, which is not available here.
- Out of scope, possible follow-ups:
  - Recover `<|tool_call>call:name{…}<tool_call|>` text into a structured call in `internal/llm`, which would make lost calls impossible regardless of the prompt.
  - Strip the empty `<|channel>thought<channel|>` markers that `gemma-4-31b-noreason` sometimes leaves in answer text. With the final prompt the probe counted none.


EvilFreelancer:
Слишком радикальные перемены, плюс проблема оригинальная с Gemma на стороне NeuralDeep и по сути данный PR её не решает, проблема так же часто вылезает. Лучше отменить и начать с решения на стороне NeuralDeep после чего вернуться к этому в Coddy.

PR #240, opened 2026-09-14
feat(session): read a follow-up written while the turn works

EvilFreelancer:
## What this is

The moment an operator knows most about what the agent should do next is while it is working — the tool call that read the wrong file, the platform nobody mentioned. Until now that was the one moment nothing could be said: the composer turn lock refuses a second prompt, so the correction had to wait for the answer or cost a **Stop**.

A message written **during** a turn is now queued on the session instead of refused, and the running turn reads it **at its next step** — between the tool calls the model just made and the request that follows them. Until the agent reads it, a queued message is still the operator's: it can be taken back.

Docs: [`docs/features/message-queue.md`](docs/features/message-queue.md).

## Where the queue is read

- **Between steps** — the ReAct loop drains the queue at the top of every iteration, before it builds the request that answers the tool results it just collected. This is the mid-turn read, and it is the point of the feature.
- **At the end** — a message written while the answer was already being composed is read by the same turn rather than waiting for the next prompt.

## The turn boundary has no window

The queue opens when a turn is admitted and is gone when it releases; nothing is persisted, and a **Stop** drops what was waiting. Draining and closing is **one** atomic step (`TakeQueuedMessagesOrClose`): either it hands back a non-empty batch and stays open, or it finds nothing and closes in the same critical section. Two operations would leave a gap in which a message is accepted by a turn that is already over — the one way a queued message could be silently lost.

## A shared session has a shared queue

A session is not owned by the tab that opened it. Every change travels down **two** paths, each carrying the whole queue and a `version`:

- the turn's own stream (so the client driving it, and anyone teed onto `composer-stream`, has it immediately);
- `GET /coddy/events`, which every browser holds open and the console subscribes to under `--remote`.

They are separate connections, so frames can arrive in either order; the version decides, and an older frame is dropped. Verified live in two real browsers on one session: what one queued appeared in the other without a reload, and a cross clicked in one removed it in both.

## Surfaces

| | |
|---|---|
| **Web** | The composer stays live. **Enter** queues; the primary control queues a draft and still **stops** an empty one. Waiting messages stack above the composer, each with its own cross. |
| **Console** | Submitting during a turn queues the prompt; the widget above the input numbers what the turn will read. `/queue`, `/queue drop <n>`, `/queue clear`; `escape` cancels the turn and the queue with it. Works under `--remote`. |
| **HTTP** | `GET`/`POST`/`DELETE /coddy/sessions/{id}/queue`, `DELETE .../queue/{message_id}`. Refusals carry a code: `no_active_turn` (409, the client sends it as an ordinary prompt instead), `queue_full`, `subagent_read_only`, `not_found`. |

## Screenshots

**Two follow-ups waiting while the agent works** — new surface, no before.

queue

**The primary control with a draft written mid-turn** — before this change the same state showed **Stop** and Enter was swallowed.

armed

**The console** — before this change the same keystroke printed `A turn is already running (escape to interrupt)`.

console

Captured against a real `coddy serve` with a stub model (Playwright, 1280 px, Dark) and through a real pty (`examples/cli/capture_queue.py`). Light theme and 390 px are untouched by this change beyond the shared tokens the cards use.

## Tests

- `features/message_queue.feature` — mid-turn read, order, taking one back, the end-of-turn read, the queue gone when the turn is over. Driven through the real manager and ReAct loop with a scripted model (`internal/agent/bdd_message_queue_test.go`). Verified red first: neutering the drain fails 3 of 5 scenarios.
- `features/message_queue_http.feature` — the routes over real HTTP, the refusal on an idle session, and **what one client queues is seen by every other client of the session** (a watcher on the turn stream and a third client on `/coddy/events`).
- `features/cli_tui.feature` — a prompt written during a turn is queued and shown, then cleared when the turn is cancelled.
- Unit: `internal/session/turn_queue_test.go` (cap, cancel, copy semantics, the atomic boundary, a concurrent writer against a concurrent drain under `-race`), `internal/remote/events_test.go` (the subscription delivers the queue, `Close` ends it, `SetServer` does not start one), `external/ui/src/ui/chat/composerQueue.test.tsx`, `serverEvents.test.ts`.

`make test` green (50 packages), `make lint` and `make lint-windows` clean, `make docs-check` clean.

## Not in this change

- **Attachments** — a queued follow-up is text; files stay in the composer for the next prompt.
- **The Telegram gateway** still answers busy rather than queueing; it has no way to show or cancel a queue.
- **Persistence** — nothing is held for an idle session or across a restart.
- The documentation layer for coddy.dev (`make site-docs`) follows the merge, since `llms.txt` names pages by their path on `main`.


EvilFreelancer:
## Cross review round

Three reviewers were given the whole diff inline with no tools. Codex (`gpt-5.6-sol`, high reasoning) and Cursor (`auto`) both answered and **independently named the same three blockers**; the third (Coddy on the local `neuraldeep/qwen3.8-27b`) produced nothing in 25 minutes, as it has before on a long brief.

Every finding below was reproduced before it was fixed.

| Finding | Verdict | What changed |
|---|---|---|
| The snapshot and its version were read under separate locks | **Confirmed** | `QueueSnapshot()` returns both under one lock; a version can no longer name content other than its own |
| The agent's own drain published straight to the turn's sender, with no version and no session id, and never reached the observers | **Confirmed** — the shared-session promise broke exactly when a message was read: a second tab kept showing it | Every mutation announces itself through one notifier the manager installs for the life of the turn; the agent publishes nothing of its own |
| REST answers carried no version, and `applyQueue` was called with two arguments where it takes three | **Confirmed** (and the missing argument poisoned the high-water mark with `NaN`) | The answers carry the version; the ordering rule moved into `chat/messageQueueState.ts` with its own tests |
| The boundary loop ran the runner again after a **cancelled** stop, and gave each continuation a fresh `max_turns` | **Confirmed** — pinned by a test that fails on the old code (`a cancelled turn ran the runner 2 times, want 1`) | Only a turn that ended with an answer continues, at most 8 times; past that the queue closes and the session is released |
| The queue was read **before** the `Stop` hooks, skipping them | **Confirmed** (Cursor) | The read moved after the hook block |
| The console applied every queue update with no ordering | **Confirmed** | The widget keeps its own high-water mark |
| A follow-up refused as busy lost the operator's text | **Confirmed** | `restoreDraftOnBusy` puts it back in the composer |
| The version restarted from zero when a session's live state was rebuilt | **Confirmed** — I had found this one myself while the reviewers ran | The counter is process-wide, and a client forgets its mark on `turn_started` |
| OpenAPI documented `400` only on POST | **Confirmed** | Documented on every queue route |
| `translateQueueError` did not map `subagent_read_only` | **Confirmed** | Mapped |
| The HTTP harness drained the queue inside its own stub, so the scenarios never touched the manager's boundary | **Confirmed** — a real hole in the tests | The stub no longer drains; the scenarios exercise the real path |
| The concurrency test proved "not read twice" but not "not lost" | **Confirmed** | It now records every accepted id and asserts each was either read or left in the queue |
| SSE test helpers could hang instead of failing at their stated timeout | **Confirmed** | One reader goroutine per stream, racing a deadline |

Not changed, with reasons:

- **A continuation is a fresh `Agent.Run`, not the same ReAct budget.** That is the intent: a follow-up is a new prompt answered inside the same admitted turn, so it gets a prompt's budget. The unbounded-work concern it raised is handled by the cap instead.
- **The permission-resume and run-plan paths do not get the boundary continuation.** They open and read a queue like any turn; what arrives in their final moments is dropped with a warning. Now documented rather than silently different.

`make test` green after the fixes and after merging `main`, `make lint` / `make lint-windows` clean, `make docs-check` clean.

PR #242, opened 2026-09-14
fix(build): the image stages every asset the SPA symlinks

EvilFreelancer:
## Why

The `Docker build and push` job has been failing on every merge since 2026-09-13. `Release binaries` keeps succeeding, so tags and archives went out as usual and only the published container stayed behind - it still serves the SPA from before the sign-in screen landed.

```
[UNRESOLVED_IMPORT] Error: Could not resolve '../../assets/coddy-logo-wordmark.svg' in src/ui/auth/SignInScreen.tsx
[UNRESOLVED_IMPORT] Error: Could not resolve '../../assets/coddy-logo-wordmark-light.svg' in src/ui/auth/SignInScreen.tsx
ERROR: process "/bin/sh -c npm run build:go" did not complete successfully: exit code: 1
```

## Cause

`external/ui/src/assets/*.svg` are symlinks into `docs/assets/`. `COPY external/ui/ ./` carries them over as symlinks, so the Node stage only sees a file if the Dockerfile also stages the target at the very path the link resolves to (`/docs/assets/`). That `COPY` named four files by hand; `feat(ui): put the wordmark on the sign-in card` (#230) added two more links and nothing brought them along.

Three further links - `coddy-logo-mark.svg`, `coddy-logo-mark-light.svg`, `coddy-logo-mark-icon.svg` - were already in the same state and would have broken the image the moment anything imported them.

## Change

- the `COPY` globs `docs/assets/coddy-logo-*.svg` instead of listing files, so a new logo needs no edit here;
- `external/ui/image_assets_test.go` walks the symlinks and fails when one resolves outside `docs/assets/` or matches no `COPY` pattern - locally, instead of in the release pipeline.

## Verification

- `docker build --target ui-builder .` - green, `npm run build:go` emits `dist/app.js` + `dist/styles.css`; the same build fails on `main`;
- `go test -tags=ui ./external/ui/` - green; reverting the `COPY` to its old form makes it fail and name all five links;
- `make lint` - 0 issues.


EvilFreelancer:
Merged into [#243](https://github.com/coddy-project/coddy-agent/pull/243) along with the rest of the run, so the whole set reviews and releases as one change. The Dockerfile glob and `external/ui/image_assets_test.go` are in there unchanged.

PR #244, opened 2026-09-14
fix(ui): a tool row reads a path against the directory the work is in

EvilFreelancer:
## The bug

A transcript row names what the call touched next to the label. That target is an absolute path whose first forty characters are the same on every row, so the ellipsis ate the part that actually tells the rows apart.

```
читаю файл /storage/Repository/coddy/coddy-agent/.coddy/worktrees/fix-session-stop-queue/do…
читаю файл /storage/Repository/coddy/coddy-agent/.coddy/worktrees/fix-session-stop-queue/do…
читаю файл /storage/Repository/coddy/coddy-agent/.coddy/worktrees/fix-session-stop-queue/DE…
```

## The change

A path target is spelt against the directory the work is happening in, when that spelling is shorter. The roots are the session's own directory followed by the **worktrees of its workspace** (`workspaceCtx.worktrees`), and the **deepest root holding the file wins** - a session opened on the checkout that runs its work inside a worktree reads against that worktree, because the path to it is the one part every row of that work repeats.

```
читаю файл DESIGN.md
читаю файл docs/nav.yaml
```

- `relativeToolTarget` (`chat/toolTargetPath.ts`) picks the deepest containing root, and otherwise falls back to the session directory and walks up only while the result stays shorter than the absolute path, so `/etc/hosts` from a four-deep session keeps its own spelling. Windows paths compare case-insensitively and keep their separator; a different drive or a UNC share has no relative spelling and is returned untouched;
- `toolCallTargetIsPath` (`chat/permissionToolPreview.ts`) gates it, so a command, a search pattern, a url, a skill or an agent name is never rewritten - a `run_command` whose command line happens to start with `/usr/bin` still reads exactly as it ran;
- the absolute path stays in the row's `title` attribute and in the expanded card;
- the live status line under the typing dots follows the same rule, through the same helper.

The roots reach the row as a new `pathRoots` prop (App to ChatScreen to MessageList to ToolCallMessage). With no roots nothing is rewritten.

## Tests

- `chat/toolTargetPath.test.ts` - 13 cases: work inside a worktree, the deepest root winning over the parent that also holds the file whatever the order, a file outside every worktree, a sibling reached by walking up, a path far enough away that absolute wins, the root itself, a relative target, a url, a command, no roots at all, Windows case folding and separators, another drive, a UNC share;
- `chat/permissionToolPreview.test.ts` - the predicate tells path arguments from commands, patterns, urls and names;
- `messages/ToolCallMessage.test.tsx` - the row renders the path relative to its worktree while the `title` keeps the absolute one, asserted exactly (a substring match would pass without the rewrite), and a command row is left alone.

## Verification

- `npx vitest run` - 168 files, 1294 tests, green;
- `make lint` - 0 issues;
- `make docs-check` - green;
- driven in the real SPA: a session rooted at the main checkout reading files inside a worktree shows `DESIGN.md` and `docs/nav.yaml`, their tooltips the absolute paths, and the `run_command` row keeps its command verbatim.

## Docs

`DESIGN.md` (the tool-call row) and `docs/surfaces/web-ui.md` record the rule, the root order and what it deliberately does not touch.


EvilFreelancer:
Merged into [#243](https://github.com/coddy-project/coddy-agent/pull/243) along with the rest of the run, including the correction that a path reads against the worktree the work is in rather than against the session directory.

PR #272, opened 2026-09-15
Tags, an archive and grouping over the stored sessions

### LINKED ISSUE #199: UI-сессии: добавить группировку по группам, тегам и проектам
## Summary

В текущем UI-списке сессий сложно быстро ориентироваться, когда их становится много. Нужна возможность группировать сессии по разным измерениям (группы, теги, проекты), чтобы ускорить навигацию и снизить «шум» в истории.

## Предложение

Добавить в интерфейсе сессий поддержку группировки/фильтрации по:

1. **Группам** (например, пользовательские или предопределённые коллекции)
2. **Тегам**
3. **Проектам** (связь с `homeDir`, репозиторием/конфигом, пр. контекстом)

В идеале это может быть сочетанием:
- режима группировки (как отображать списки),
- фильтров по одному/нескольким значениям,
- возможности быстро сворачивать/разворачивать группы.

## Текущее поведение

- Все сессии идут в один плоский список.
- При большом количестве сессий сложнее быстро выделить нужную область.

## Ожидаемое поведение

- В UI добавить переключатель способа группировки: по времени / по проекту / по тегам / по группам.
- На уровне сессий доступны теги и группа/проект, по которым можно группировать и фильтровать.
- Пользователь может быстро перемещаться между группами без ручного скролла.
- Настройки группировки сохраняются между открытиями (в рамках пользователя/профиля, если есть состояние).

## Acceptance Criteria

- Пользователь может группировать список сессий хотя бы по одному из режимов: **проект / тег / группа**.
- Группы корректно отображаются и обновляются после создания/изменения сессии.
- Есть возможность фильтровать/показывать только выбранные группы/теги.
- Поведение не ломает существующие операции со сессиями (удаление, открытие, экспорт и т.д.).

## Дополнительно

Если у сессий уже есть метаданные, пригодные для группировки, лучше переиспользовать их, минимально расширяя модель данных. Если нет — добавить минимально-инвазивные поля/атрибуты.

EvilFreelancer:
Closes #199.

A history grown over months is narrowed rather than scrolled. Three fields join the session bundle, and the two surfaces that read it grew the controls to use them.

## The model

`session.json` gains three things:

- **`tags`** - a short flat vocabulary a conversation is filed under, normalized on the way in (lower case, inner whitespace as a hyphen, duplicates dropped, at most 8 of at most 32 characters);
- **`archived`** / **`archivedAt`** - a conversation put aside without being taken off disk;
- **`origin`** - the surface that started the session, empty for one opened on this host and `gateway:telegram` for a chat the messenger gateway is holding. It is written once by the surface that creates the session, so reopening a Telegram chat from the browser does not relabel it.

Nobody has to invent the tags. `POST /coddy/describe`, the call that already names a new chat, asks the model for the phrase **and** one to three labels in the same answer, and the `PATCH` that pins the title carries them along. Filing costs no second request, and a model that ignores the instruction names the chat exactly as it did before.

## The listing

`GET /coddy/sessions` takes `archived` (`exclude` / `only` / `all`), `tags` (any of them), `origin` (`local` / `gateway`), and `sort` (`updated` / `created` / `title` / `messages` / `tokens`) with `order`. The order applies to the whole filtered listing **before** paging, so page two of a sorted listing continues page one; a session the column says nothing about sorts last in both directions, and ties break by id so paging stays stable. `tokens` reads each bundle's `stats.json` and only when it is the column asked for. An unknown value of any of them is a `400` rather than a silent default.

Search now matches a tag beside the title, the workspace and the first user message. `bulk-delete` gained `scope: "archived"`, and `scope: "all"` now reaches into the archive too - a scope that left sessions behind because they were put aside would not be the whole history.

## History

Everything that decides what the drawer shows moved behind one control in its head: **Status**, **Environment**, **Group by**, **Sort by**. Grouping is a client concern, so switching costs no request and never reorders what the server sorted inside a group.

History grouped by folder

*Grouped by folder: the heading is the folder name with its row count, and hovering it offers a new chat in that workspace - the pick goes through the same pre-session path as the composer's folder chip, so the server resolves the folder's current git branch.*

The History filter menu

*Status leads, because "am I looking at the archive" is the question asked most often. Environment narrows by where a conversation came from and lists the configured remotes to switch to. Sections are flat rather than nested submenus: a drawer is narrow, and a hover-to-open submenu in it is a worse target than one more line of text.*

Each row gained an archive tray beside its trash, and its tags render under the title.

## The session table in Settings

The session table

*Everything stored, the archive included.*

- the **duplicate per-row delete is gone**: the tick and the one toolbar button are the whole destructive surface, so the scope of a click is never ambiguous;
- **column headers sort** the whole listing, with `aria-sort` and a caret on the sorted one;
- an **archive filter** beside the search, and archived rows marked as such;
- **one button empties the archive**, behind the shared confirmation dialog, which says in words that it reaches past the page;
- **tags are chips you can press** to filter the table.

## Tests

- `features/session_tags_archive.feature` - seven happy-path scenarios over the live HTTP surface (godog, stub model), covering tags, the tag filter, the archive, emptying it, ordering, the tags describe proposes, and telling a gateway chat apart from one started here;
- unit tests for the edges: normalization and its caps, sorting rules, the query values that must be refused, clearing tags with an empty array, the archive stamp, the origin that cannot be overwritten, and the describe parsing that still names a chat when the model ignores tags;
- vitest for the new pure helpers (grouping, the tag vocabulary, the cookie) and for both surfaces.

`make test` green, `make lint` clean, `make docs-check` clean; vitest 1391 passed.

Documentation: `docs/features/sessions.md` (tags and the archive), `docs/surfaces/web-ui.md` (both surfaces, three screenshots), `docs/reference/http-api.md`, `DESIGN.md` and the served OpenAPI.


EvilFreelancer:
## Cross review

Three reviewers over the diff: **Cursor** (auto) on everything, **gpt-6-astra** on the Go side, **gpt-5.6-sol** on the SPA, and **qwen3.8-27b** on the Go side as a fourth opinion. Nine findings confirmed and fixed in `9d26f5d5`, each with a test that fails without the fix. Two were checked and did not hold.

### Fixed

| What | Where |
|---|---|
| **Panic** in the describe handler on a tag line whose lower case is longer than the original (`Ⱥ` is two bytes, `ⱥ` is three, so the offset ran off the front of the string) | `coddy_coddy.go` |
| **Non-transitive comparator**: a chronological order for readable stamps mixed with a lexical one for the rest leaves `sort.SliceStable` entitled to anything | `list_query.go` |
| **Tag filter capped at eight** alternatives by the per-session storage limit | `list_query.go` |
| **A tag holding a comma** could be stored and never asked for again | `list_query.go` |
| **Archive flag and its stamp read under separate locks**, so a save racing a change could persist half of it | `state.go`, `filesystem.go` |
| **Emptying the archive deleted the open conversation**, walking past the table's own promise | `SessionsManager.tsx` |
| **Nested state updaters** in the sort handler: StrictMode ran the inner one twice, so clicking the sorted column flipped it back | `SessionsManager.tsx` |
| **Optimistic archive** moved the row before the server agreed, and two requests for one row could be in flight at once | `App.tsx` |
| **Age buckets subtracted 24 hours** where they meant a calendar day (23 or 25 hours around a DST change) | `sessionGroups.ts` |

Plus one I found myself before the reviewers: the `+` on a folder heading posted the workspace to the conversation being **left**, and the new chat never got it. Leaving a session is asynchronous, so the pick is parked and applied by the one effect that owns the session change.

### Checked, did not hold

- *"Switching to a remote keeps the origin filter"* (Cursor and sol, independently) - `connectRemote` ends in `window.location.reload()`; no client state survives the switch.
- *"The decorated tag line test fails"* (qwen, reported as blocking) - `describeStripLineNoise` strips the `**` pair before the prefix is matched; the case passes.

### Known and left

The title suggestion replaces tags wholesale, so a manual edit made while `/coddy/describe` is still in flight would lose to the model's proposal. There is no tag-editing surface in the SPA yet, so the window is not reachable today; it wants a conditional update when one is added.

`make test` green, `make lint` clean, `make docs-check` clean, vitest 1399 passed.

PR #277, opened 2026-09-15
The model files its own session, and so can you

EvilFreelancer:
Two halves of the same thing: the model can file the session it works in, and the operator can file it by hand.

## The tool

`session_describe` reads how the current session is filed and writes the parts a call names.

| Argument | What it does |
|---|---|
| none | reports the current title and tags, changes nothing |
| `title` | renames the session; an empty string drops the pin and gives back the title derived from the first message |
| `tags` | replaces the whole set; an empty array clears it |
| `add_tags` / `remove_tags` | change the set in place, leaving the rest alone |

`tags` together with `add_tags` or `remove_tags` is refused: they are two different intentions, and a call that carries both leaves the order between them to whoever reads the code. Removals are matched after the same folding the storage uses, so a caller may name a label the way it reads ("Session Store") rather than the way it is stored (`session-store`), and the answer reports the stored spelling, which is what a filter has to be asked for. Additions past the per-session cap are what gets dropped, so filing one more topic cannot unfile the eight already there.

The layers stay where they were: the tool parses arguments and reports what moved, `internal/agent/session_filing.go` decides what an update means, `internal/session/filing.go` owns the folding. Agent and plan mode may file; ask may not, because it writes.

## Editing the same thing by hand

One editor, two entry points. History reaches it from the row menu, which gains **Rename** and **Tags** above the rule; the session table from a dashed `+` after the chips, where the chips themselves stay filters.

The row menu

*Pin, rename and tags change what the row says about itself; the rule still opens the pair that takes the conversation out of the list*

The tag editor with a label being typed

*A cross on every chip, the labels this history already uses under the box, and the folded spelling shown only when it differs from what was typed*

The label filed

*There is no Save: the change is a PATCH at once, and the row behind the editor takes the set the server answered with*

Renaming a row in place

*Rename turns the title into the box the chat header already uses, with the name selected; Enter saves, Escape leaves it alone*

The session table

*Before: the tags were chips you could filter by. After: a dashed + at the end of them, visible at rest because a finger has no hover*

The editor from the table

*The + opens the same editor, anchored under itself*

The editor in the light theme

*Light theme: the editor is cut from the tooltip surface, like the menu it opens from*

## Two things behind it

**Every first message is described now.** `POST /coddy/describe` used to echo an input of three words or fewer and propose nothing, which was cheap while a title was all it produced. The tags ride on that call, so the shortest conversations were the only unlabelled ones - `git status` is a usable title and no filing at all.

**That call's PATCH is a suggestion, not a rename.** It carries `titleIfUnpinned`, so a session already named during the first turn - in the chat header, or by the agent through `session_describe` - keeps its name, while the tags of the same answer still land. Without it the describe answer, in flight since before the turn, would overwrite whatever the turn decided to call the session.

## Tests

- `features/session_describe_tool.feature` (5 scenarios) drives a real `Agent.Run` with a stub provider over a real bundle on disk, and checks both what the tool reported and what `session.json` holds afterwards;
- `features/session_tags_archive.feature` gains the short first message that is described and filed;
- unit tests for the folding (`MergeTags`, `NormalizeTitle`, the title limit), for what the tool refuses, for what `applySessionFiling` writes and leaves alone, for `titleIfUnpinned`, and for the editor's arithmetic (`tagEditing.ts`);
- `make test` green, `make lint` clean, `npx vitest run` 1473 tests green.

Verified live as well: a real model called the tool through `coddy -p` and the folded tags landed in `session.json`, and the tag editor was driven in a browser against `coddy serve` for the screenshots above.


EvilFreelancer:
## Cross review

Three independent reviewers over the full diff: `coddy` on `neuraldeep/qwen3.8-27b`, `coddy` on `codex/gpt-5.6-sol` (high), and `cursor-agent` on `auto`. Every finding was verified here before it was acted on.

**Confirmed and fixed** (commits `82584304` and `9fa31dbe`):

| What | Who found it |
|---|---|
| `titleIfUnpinned` was a read followed by a write, so it still had the race it exists to close | sol, qwen |
| `add_tags` merged outside the session, so a label filed from another surface between the read and the write was dropped | sol |
| `changed` was derived by comparing the effective title before and after, so clearing a pin that read like the derived title was reported as no change | sol |
| `PATCH /coddy/sessions/{id}` trimmed a title while the tool folded and limited it - one field, two vocabularies | sol, qwen |
| Two tag gestures made before the first `PATCH` answered both started from the set before both, so the second undid the first | cursor, sol |
| An inline rename could be saved twice (Enter then blur) or saved after Escape discarded it | cursor |
| History swallowed a refused tag write, and a dropped connection left the promise rejected | cursor, sol |
| The local fold used JavaScript's `\s`, which carries U+FEFF the server keeps inside a label and misses U+0085 the server treats as a space | sol |

The two races now have tests that run eight goroutines at them, and the three UI defects have tests of their own.

**Checked and not defects.** Both `coddy` reviewers read the Go layer as a separate brief and reported `words` as unused in `coddyDescribePost` after the short-circuit was removed; it is still read twice below the hunk they were given (`describePickPhraseFromLLM` and the fallback), and the package builds. One reported the tool as undocumented and one reported the new screenshot as missing from the diff: both are in the change, in `docs/reference/tools.md` and `docs/features/sessions.md`, and as a binary file the text diff does not carry.

PR #317, opened 2026-09-19
feat(tools): preview_server serves project files on a free localhost port

hijera:
## Why

HTML, JS and CSS work often cannot be opened from disk: a page loaded as `file://` cannot import ES modules, `fetch` its own data or start a worker. Codex and other coding agents can put the project on a local web server; Coddy had to hope for a `python` or an `npx` on the machine. Now "run this on a web server, I want to click around myself" gets a link.

## What

- **`preview_server` tool** (`internal/tools/preview`): serves a directory of the session's working directory on `127.0.0.1:<free port>`, returns the address and tells the model to give it to the user and invite them to open it. Args: `path` (a directory, or a file whose directory is served and which the address opens) and an optional `timeout_seconds`.
- **Runs on goroutines inside coddy**, as a task of the existing pool: new `bgtask.KindServer`, `Spec.URL` / `Snapshot.URL` and `Spec.NoTimeout`. So it shows among the running tasks, `background_list` prints its address, `background_output` is its request log (`GET /js/app.js 404` - the usual answer to "the page is blank"), `background_stop` / the Stop button / deleting the session / drain end it. It has no hard timeout unless the call names one; that one is honoured verbatim, past `tools.background.max_timeout_seconds`.
- **Made for editing**: files are read on every request with `Cache-Control: no-store`; `.js` / `.mjs` always go out as `text/javascript` (the Windows registry often says `text/plain`, which breaks ES modules). Asking again for a directory that is already served returns the address that is up.
- **Narrow on purpose, because it runs without a permission prompt**: only the working directory or below (symlinks resolved, files opened through `os.Root`), dot-files (`.git`, `.env`, `.coddy`) answer 404, `GET`/`HEAD` only, a foreign `Host` is refused on a loopback bind (DNS rebinding), no CORS headers. Agent mode only.
- **Config** `tools.preview_server`: `enable`, `host` (default `127.0.0.1`), `public_host` (the host written into the address, for Docker or a remote machine) - struct, embedded schema, `config.example.yaml`, settings form, `configure-coddy` skill, generated reference. `tools.background.enable: false` hides the tool as well.
- **Surfaces**: the Tasks panel tags the card `server` and shows the address as a link that opens in a new tab (above the card's stretched opener, like Stop), the open card has the address with a copy control; `/tasks` in the console shows the same; en + ru dictionaries; OpenAPI and `docs/reference/http-api.md` describe `kind: server` and `url`.
- **Docs**: new `docs/features/preview-server.md` (with a capture), `nav.yaml`, the tools reference, a section in background tasks, README, the agent prompt.

## Tests

- `features/preview_server.feature` - 8 scenarios against a real pool, a real directory and a real HTTP client: the address on a free port, the invitation to the user, the task in the listing, an edit served fresh, the same address on a second call, stop, a timeout, the session deleted.
- Unit tests for the edges: outside the working directory, a missing directory, a file argument, dot-files, a symlink out of the root, a foreign `Host`, script content types, read-only, the request log, the port released when the pool refuses the task, both switches, config validation, mode tool sets; `internal/bgtask` for the untimed task; vitest for the card; the console's tag and title.
- `examples/httpserver/http_e2e_preview_server.py` (added to `test_httpserver.sh`): run against `neuraldeep/qwen3.8-27b` - asked in an operator's words, the model called `preview_server`, the answer carried the address, the page and its ES module were served, `.env` was not, the edit showed, stop took the address down. The same with the Russian phrasing from the request ("запусти мне это на вебсервере, хочу сам потыкать").
- Verified in the running build: the demo page with ES modules loads and works from the address the agent handed out, and the port is released when coddy exits.
- `golangci-lint`: 0 issues on all three passes; `tsc --noEmit` clean; generated docs regenerated (from an LF export, this checkout is CRLF).
- Local `go test` with all tags: the packages this change touches are green. The Windows box I ran on fails a handful of tests in untouched packages for environmental reasons (PowerShell exit codes in the wake scenarios, `~` with a redirected HOME, CRLF in the embedded docs and in two stylesheet tests, timeouts under load), and it has no cgo, so **the race detector did not run locally** - one test that read the request log right after the response was made to wait for the line instead, since the handler writes it after the client has its answer.

## Screenshots

The Tasks panel is the only changed surface; the card kind is new, so there is no "before".

| | |
|---|---|
| Running server, folded card (dark, 1280) | Open card: address with copy, request log (dark, 1280) |
| folded | open |
| Light (1280) | Russian interface (dark, 1280) |
| light | ru |

Phone width (390):

## Not in this PR

`make site-schema` / `make site-docs` write into the site repository; the change only adds an optional key, so the schema is safe to publish right away, and `llms.txt` follows the merge.


EvilFreelancer:
## Cross-review round + follow-up fixes (d2c6ac00)

Ran this PR through four independent reviewers (cursor-agent, codex `review`, devin, and a fresh-context subagent) plus a manual pass. Findings below are consolidated; the follow-up commit d2c6ac00 fixes them and adds ~13 edge tests on top of the original suite.

### Fixed (P1)

- **`tools.preview_server` was dropped by the JSON config DTO** — `ConfigToJSONDTO` / `JSONDTOToConfig` had no field for it, so any unrelated **Save all** in Settings silently reset `enable: false` and custom hosts. Added `ToolPreviewServerJSON` + both conversions + a round-trip test.
- **Dot-file policy had holes**: `path: ".git"` (or a dir under a dot-named one) served the whole dot-tree at clean URLs like `/config`; a `file.txt -> .env` symlink inside the root served the secret; directory listings showed dot names. Now: a dot-named root or file arg is refused in `resolveRoot`, `dotHiddenFS` filters dot entries from `ReadDir`, and `hiddenTarget` resolves the request path via `EvalSymlinks` and refuses dot-named targets.

### Fixed (P2)

- **Host allowlist was decided by the bind-host string** — `host: LOCALHOST`, `localhost.`, `ip6-localhost` bind loopback but parse as non-IP, which silently turned the DNS-rebinding allowlist off. Loopback is now read from the bound listener's `TCPAddr.IP.IsLoopback()`.
- **Dedup reused a stale server** — `sameDir` compared pathnames, so a build that deletes and recreates the served directory kept answering from the deleted inode (500s). The check now compares the opened directory's identity (`root.Stat(".")` + `os.SameFile`), stops the stale server, and starts a fresh one; the check-and-launch section is serialized (`previewMu`) so two parallel calls can't bind two ports.
- **`timeout_seconds` could wrap `time.Duration`** — a named limit past ~292 years overflowed into *no* timeout. Now clamped to `maxClockTimeoutSeconds` on both resolution paths.
- **`[::1]` as `host`** validated, then broke `net.JoinHostPort` (`[[::1]]:0`) — brackets are rejected in `validate`.
- `server.Stop` is nil-safe before `serve()` and closes `done` so `Wait()` can't hang; `background_list` stops printing the dead URL once the task finishes; `AGENTS.md` synced with the rules trees (rules-sync was missing); `preview-server.md` / `background-tasks.md` updated (listing/symlink coverage, "not password-protected", orphaned rows are cleared not stopped, e2e script listed).

### Fixed in review-of-the-fix (subagent pass on the follow-up itself)

- `liveServers` was keyed by bare task id — task ids are **per-session**, so session B's `bg_1` overwrote session A's and the wrong server got stopped / evicted. Now keyed `sessionID + "/" + taskID`.
- `dotHiddenDir` wrapped every file and stripped `io.Seeker` — `Range:` requests (video/audio) would have failed. Only directories are wrapped now.
- `hiddenTarget` skipped its check when the served root path contained a symlink (e.g. a `t.TempDir` under a linked `/tmp`) — root is resolved once in `newHandler`.
- Dedup `Load`-miss reused a possibly-dying server — now falls through to a fresh launch.

### New tests (all green)

dot-named dir/file refusals, dot names absent from listings, symlink-to-dotfile 404, `LOCALHOST`/`localhost.` bind keeps the allowlist, recreated-directory restart, parallel same-dir calls → one server, cross-session dedup, `Range:` request, HEAD, stop-before-serve no panic, fresh port after stop, JSON DTO round-trip, timeout clamp, bracketed host rejected.

### Verification

- `make lint`: 0 issues on all three passes + `tsc --noEmit` clean; `gofmt` clean.
- `make docs-check`: ok.
- `make test`: everything green here. (One unrelated local flake in `TestCoddyConfigPutConnectsMCPToActiveSession` caused by a `~/.coddy/mcp.json` on this machine leaking a playwright server into the test session — passes with an isolated `CODDY_HOME`.)

Left intentionally as-is: an orphaned `kind: server` record after restart still shows its dead URL (documented — cleared, not stopped); `public_host` on a loopback bind widens the allowlist on purpose (tunnel use case, documented); `hiddenTarget` has an inherent check-then-open race (os.Root still confines regardless).

